PR #5, opened 2018-07-10
Added 3 level paths and added a test. Fixed -d flag.

ztulloch:


amouat:
This looks good! It definitely doesn't matter for this, but normally it would probably be best to have the `-d` fix in a different PR.

hamhut1066:
Thanks for the changes, I left some feedback on the PR

amouat:
Thanks Hamish! Yeah, good point about "?" and the printlns. Zander, I think you can just push an update to the branch for this PR, then I think it's pretty much good to go.

ztulloch:
Thanks for the feedback Hamish! I'll sort those out.

amouat:
Removed unwraps and merged. We still need to take care of testing getting manifests with different levels (name vs user/name vs org/user/name) and refactoring.

PR #6, opened 2018-07-31
Manifest changes

ztulloch:
Added function for onename manifest and added tests for 1/2/3 level manifests.


hamhut1066:
Looks good to me

PR #7, opened 2018-08-24
Updates to build around new rocket version. Removed !use_extern_macros which has gone stable.

ztulloch:


amouat:
Looks good. You don't need to note the previous version as it's in git anyway. I shouldn't have left the old commented out stuff.

PR #9, opened 2019-02-18
Basic Authentication

ztulloch:
Implementation of a very basic authentication system that talks to docker using a hardcoded user/password combination.


src/response/errors.rs
```diff
@@ -47,8 +48,10 @@ struct ErrorMsg {
 //This needs refactored.
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        debug!("error formatter");
```
amouat: Think we can delete this one

src/response/errors.rs
```diff
@@ -73,8 +76,10 @@ impl fmt::Display for Error {
 
 impl error::Error for Error {
     fn description(&self) -> &str {
+        debug!("Error matcher");
```
amouat: Delete this too

src/response/html.rs
```diff
@@ -8,6 +8,7 @@ pub struct HTML<'a>(pub &'a str);
 
 impl<'a> Responder<'a> for HTML<'a> {
     fn respond_to(self, _: &Request) -> Result<Response<'a>, Status> {
+        debug!("html response");
```
amouat: Delete this too

src/response/repo_catalog.rs
```diff
@@ -11,10 +11,11 @@ impl<'r> Responder<'r> for RepoCatalog {
     fn respond_to(self, _req: &Request) -> response::Result<'r> {
 
         let json = serde_json::to_string(&self).unwrap();
+        debug!("repo catalog response");
```
amouat: Delete this

src/response/trowtoken.rs
```diff
@@ -0,0 +1,100 @@
...
+use std::time::{SystemTime, UNIX_EPOCH};
+use rocket::http::{Status};
+use rocket::http::{ContentType};
+use rocket::request::Request;
+use rocket::response::{Responder, Response};
+use crypto::sha2::Sha256;
+use chrono::Local;
+use jwt::{Header as TokenHeader, Token};
+
+const AUTHORISATION_SECRET: &str = "Bob Marley Rastafaria";
```
amouat: 😂 I worry about you sometimes

src/response/trowtoken.rs
```diff
@@ -0,0 +1,100 @@
...
+
+#[derive(Debug, Serialize)]
+pub struct TrowToken;
+
+#[derive(Debug, Serialize, RustcEncodable, RustcDecodable)]
+pub struct BearerToken {
+    user_id: String,
+    client_id: String,
+    scope: String,
+    iat: u64,
```
amouat: what's iat? I'm sure it's from a spec, but can we put it in a comment?

src/response/trowtoken.rs
```diff
@@ -0,0 +1,100 @@
...
+}
+
+fn encode_token() -> Result<String, Error> {
+    // hard coded values for token
+    let username = "admin";
+    let client_id = "docker";
+    let scope = "push/pull";
+    let now = SystemTime::now();
+    let current_time = now.duration_since(UNIX_EPOCH).expect("Time went backwards");
+    let expiry_time = 3600;
```
amouat: Can we add a comment with what this is in seconds?

src/response/trowtoken.rs
```diff
@@ -0,0 +1,100 @@
...
+    let token_enum = bearer_token.signed(AUTHORISATION_SECRET.as_bytes(), Sha256::new()).ok();
+    let mut token_string = String::new();
+    /*
+    match token_enum {
+        Some(token_enum) => token_string = token_enum,
+        _ => (),
+    };
+     */
+    if let Some(token_enum) = token_enum { token_string = token_enum };
+    Ok(token_string)
```
amouat: Shouldn't we return an Error rather than Ok and the empty string? Or is there valid cases where it will be empty?

src/response/trowtoken.rs
```diff
@@ -0,0 +1,100 @@
...
+}
+
+impl<'r> Responder<'r> for TrowToken {
+    fn respond_to(self, _: &Request) -> Result<Response<'r>, Status> {
+        let current_time = SystemTime::now().duration_since(UNIX_EPOCH).expect("Time went backwards");
+        let expiry_time=3600;
```
amouat: This has appeared twice, maybe have a default expiry const?

src/response/trowtoken.rs
```diff
@@ -0,0 +1,100 @@
...
+        // create token string and put it in response header
+        let token_string = encode_token();
+        match token_string {
+            Ok(token_string) => {
+                let date = Local::now();
+                let formatted_date =  date.format("%Y-%m-%dT%H:%M:%SZ");
+                let formatted_string=format!("{{\"token\":\"{}\",\"expires_in\":{},\"issued_at\":\"{}\"}}/n", token_string, expiry_time, formatted_date/*"2019-02-18T13:01:21.241377823Z".to_string() current_time.as_secs()*/);
+                println!("formatted estring is {:?}", formatted_string);
```
amouat: probably don't want this on stdout (should really be using log, but I don't think we need this stuff except in dev)

src/response/trowtoken.rs
```diff
@@ -0,0 +1,100 @@
...
+                let formatted_body = Cursor::new(formatted_string);
+                println!("formatted body is {:?}", formatted_body);
+                Response::build()
+                    .status(Status::Ok)
+                    .header(ContentType::JSON)
+                    .sized_body(formatted_body)
+                    .ok()
+            }
+            _ => {
+                println!("CATCHALL!");
```
amouat: Is this expected to happen for unauthorized users or should it never occur? Maybe put it in a comment and change to log message

src/response/trowtoken.rs
```diff
@@ -0,0 +1,100 @@
...
+                Response::build().status(Status::Unauthorized).ok()
+            }
+        }
+    }
+}
+
+/*
+#[cfg(test)]
```
amouat: Tests would be nice :)

src/routes.rs
```diff
@@ -2,20 +2,26 @@ use std::str;
...
 use rocket::request::{self, FromRequest, Request};
 use rocket::{self, Outcome};
 use rocket_contrib::json::Json;
 use TrowConfig;
-
+use serde_json::Value;
+use crypto::sha2::Sha256;
+use jwt::{Header, Token, Registered};
 use types::*;
+const AUTHORISATION_SECRET: &str = "Bob Marley Rastafaria";
```
amouat: Yeah, we're going to have add the secret stuff as state to Rocket so you can retrieve it from any request

src/routes.rs
```diff
@@ -58,23 +64,97 @@ pub fn catchers() -> Vec<rocket::Catcher> {
     catchers![not_found]
 }
 
-struct AuthorisedUser(String);
+struct AuthorisedUser {
+    // other useful fields could include organisation home scope permissions
+    username: String,
+    authorized: bool,
```
amouat: I'll have to double check the docs, but I think AuthorizedUser should only be possible for a user that passes authorization (i.e. this should always be true).

src/routes.rs
```diff
@@ -58,23 +64,97 @@ pub fn catchers() -> Vec<rocket::Catcher> {
...
+}
+
 impl<'a, 'r> FromRequest<'a, 'r> for AuthorisedUser {
```
amouat: I think we should move this somewhere else

src/routes.rs
```diff
@@ -58,23 +64,97 @@ pub fn catchers() -> Vec<rocket::Catcher> {
...
-        Outcome::Success(AuthorisedUser("test".to_owned()))
-    }
+        // Look in headers for an Authorization header
+        let keys: Vec<_> = _req.headers().get("Authorization").collect();
+        if keys.len()!=1 { // no key return false in auth structure
+            let auth_user = AuthorisedUser {
+                username: "".to_string(),
+                authorized: false,
+            };
+            return Outcome::Success(auth_user);
```
amouat: Hmm, I was expecting a fail here

src/response/errors.rs
ztulloch: Oops. Was just trying to work out when they got called. Should have done a grep on debug!

src/response/trowtoken.rs
ztulloch: I put this secret in twice where it needs to decode and encode the token. I need to work out the best way of having it somewhere once in Rust! hmmm I wonder what music I was listening to that day.

src/response/trowtoken.rs
ztulloch: Issued at. Thought everyone knew that! ;) Could change them to issue_time and expiry_time to make it more readable? I've been Javaing so long it's just great to use snake case!

src/response/trowtoken.rs
amouat: Yeah, I mention it below as well. I actually know how to solve this one. It's one of those things that's confusing at first, but makes sense later. You can look up state in the guide if you want to know more, but basically we'll just tell rocket to manage some struct with our secrets.

src/response/trowtoken.rs
ztulloch: At the moment it is just a meaningless number! The issue time and expiry time aren't implemented. I ended up having to use the chrono crate to create a date formatted string that Docker would understand. At this stage, I just wanted to create a token and pass it back to Docker, then have Docker connect with the token and to verify it on trow. To do more advanced token checking, I may have to switch token crates as the main one seems to be very basic.

src/response/trowtoken.rs
ztulloch: 👍

src/response/trowtoken.rs
ztulloch: I'm terrible for diagnostics! We once put out a major software release that if you pressed a certain sequence of keys it would swear at you profusely. Had to release a patch the same week! I always grepped for all major swear words before packaging up a release after that!

src/routes.rs
ztulloch: Yeah this is the bit I wasn't happy with. I couldn't work out how to get the data from the request guard to the request, so I used the auth_user like a state machine. username + true = authorised user. "bearer" + true = valid token. Anything else = unauthorized.

src/response/trowtoken.rs
amouat: I'll close this for the minute and we'll deal with it in a future PR

src/response/trowtoken.rs
amouat: Yeah, I think issued_at

src/routes.rs
amouat: Let's do that in a separate PR

amouat:
Ok, still plenty of work to do, but it compiles and passes tests so let's merge for the moment.

PR #13, opened 2019-09-30
Remove unnecessary sudo in configure-host.sh

nebril:
`cert_file` is created by user who runs the install script, so `sudo` in front of `tee` is unnecessary.


amouat:
Looks good, thanks!

PR #18, opened 2019-10-14
Fix link to main install guide

ensonic:
INSTALL.md is in the parent directory.


amouat:
That was quick, I only just moved it! Thanks :)

PR #23, opened 2019-11-08
Do not hardcode cluster name for dig

### LINKED ISSUE #16: Cluster name should be configurable
The install scripts assume the cluster name is "cluster.local". This should be automatically discovered or configurable.

All names used in install should be configurable or discovered, with reasonable defaults.

See also https://github.com/ContainerSolutions/trow/issues/11#issuecomment-541214213

lukaszo:
Instead use search list defined in /etc/resolv.conf

Closes #16


amouat:
Nice, thanks a lot! 

I have a feeling that I've taken shortcuts in other places that may mean #16 isn't quite closed yet.

PR #25, opened 2019-11-24
Update MANUAL_INSTALL.md

ianmiell:
trow.yaml appears to be in the `quick-install` folder, not `install`


amouat:
Thanks Ian. I moved a lot of stuff around last week. 

TBH those instructions aren't very useful now; for "proper" installs you're better off looking at the kustomize yaml under `install`.

PR #26, opened 2019-11-28
remove kube-public from init script, use pod namespace

sjpotter:
this is needed to allow trow to run in alternate namespaces (or even multiple)

Currently, trow;init fails as it uses $POD_NAMESPACE inconsistently and has kube-public hard coded in multiple places


amouat:
Looks good, thanks a lot! 

There was a patch to fix this merged before, but it looks like it was only half-complete, I'm not sure why.

sjpotter:
a bit more work is need to "templatize" the yaml files (service accounts et al), to really reflect this

amouat:
Yes, I'm not surprised. There should be a test for this stuff.

I'm working on another part of the code at the minute. I'm open to pull requests, but I'll create an issue in the mean time.

PR #28, opened 2019-12-01
make copy-certs work with namespaces other than kube-public

sjpotter:
continuing my efforts to ensure that trow can work in namespaces other than kube-public, this enables the copy-certs containers to work

### DIFF
--- a/quick-install/copy-certs-tmpl.yaml
+++ b/quick-install/copy-certs-tmpl.yaml
@@ -9,7 +9,6 @@
 kind: Job
 apiVersion: batch/v1
 metadata:
-  namespace: kube-public
   name: copy-certs-{{.metadata.uid}}
 spec:
   template:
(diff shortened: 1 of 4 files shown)


sjpotter:
digging into this, I dont think it needs to create configmap.  The ca.crt is part of the serviceaccount token that is part of all pods.

instead of using a configmap, can just copy

/run/secrets/kubernetes.io/serviceaccount/ca.crt

sjpotter:
note: this breaks setting up the validator webhook as it needs the same contents.  easy enough to fix with the kubectl call, and without the configmap, but wondering if there's a way to do it without the kubectl call (to make it orchestratable without a kubectl dependency)

amouat:
Sorry for the delay, I'll try to go through this all this week.

You're right about the `ca.crt`, but I do sort of like having it in an simple place. It would be great if there was an easy API way to get the cert, but there is a long issue on this and the maintainers are against it :( https://github.com/kubernetes/kubernetes/issues/61572

I would like to get rid of `kubectl` as it's going to cause incompatibilities with versions of k8s.

amouat:
Thanks a lot for this, I'm merging your PRs. Because the deploy system is a bit of a mess (we should automatically create new images for branches), I think it's easiest to merge then fix on master atm. I'll need to come up with a better solution later.

amouat:
hmm, this broke more than I expected :)

I'll have to be a bit more careful

PR #29, opened 2019-12-01
more changes to namespace usage

sjpotter:
after this, the only thing that isn't caught is the master install.sh
    
this adds templating to trow.yaml (unsure why trow-gke.yaml is needed, as I'm using trow.yaml on gke without a problem)
    
this adds an install-trow.sh that uses sed to replace {{namespace}} with proper namespace


amouat:
I suspect trow-gke.yaml was checked in by accident :)

PR #38, opened 2020-02-04
Ability to select namespace during quick install

### LINKED ISSUE #27: Allow setting of namespace in quick install
The user should be able to choose which namespace to install Trow to in the quick install (standard install can already be easily changed).

Fix should include some form of test.

mcwienczek:
Fixes #27 when done


quick-install/configure-host.sh
```diff
@@ -2,7 +2,18 @@
 set -eo pipefail
 unset CDPATH
 
-registry_host="trow.kube-public"
+namespace='kube-public'
+if [ ! -z "$1" ]
```
amouat: I suppose we should use proper argument handling now, just to avoid bugs later where the arguments get confused e.g. 
https://stackoverflow.com/questions/192249/how-do-i-parse-command-line-arguments-in-bash#13359121

quick-install/copy-certs.sh
```diff
@@ -3,16 +3,22 @@ set -eo pipefail
...
+then
+	namespace=$1
+fi
+
 echo
 echo "Copying certs to nodes"
 
 #delete any old jobs
-for job in $(kubectl get jobs -n kube-public -o go-template --template '{{range .items}}{{.metadata.name}}
+for job in $(kubectl get jobs -n $namespace -o go-template --template '{{range .items}}{{.metadata.name}}
```
amouat: I *think* we should put `$namespace` in double quotes. I guess there is as a potential security hole if an attacker can set the namespace to something like `bla; kubectl delete...`, but I don't think it's worth worrying about that one.

quick-install/copy-certs.sh
```diff
@@ -3,16 +3,22 @@ set -eo pipefail
...
 
 {{end}}') # blank line is important
 do
   if [[ $job = copy-certs* ]]; then
-    kubectl delete -n kube-public job "$job"
+    kubectl delete -n $namespace job "$job"
```
amouat: Again I think `"$namespace"`

quick-install/install.sh
```diff
@@ -46,25 +47,32 @@ if [[ "$(uname -s)" = "Darwin" ]]; then
   on_mac=true
 fi
 
+namespace='kube-public'
+if [ ! -z "$1" ]
+then
+	namespace=$1
+fi
+
+echo "Installing trow in namespace: $namespace"
```
amouat: trow -> Trow

quick-install/nginx-deployment.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: apps/v1 # for versions before 1.9.0 use apps/v1beta2
```
amouat: Can we delete the comment? I think it will cause confusion as 1.9 is so old it shouldn't be being used any more.

amouat: Also, I wonder if we're better off using a one-line kubectl command? This basically just does:
```
kubectl create deploy trow-test --image=trow.default/test/nginx:apline
```

mcwienczek: I will delete this file, committed by accident

quick-install/copy-certs.sh
mcwienczek: I think we can assume that cluster admins will install trow in the cluster. Then this issue is degraded from security issue to nice-to-have :) Nevertheless, I will add double quotes, that's a good practice indeed.

quick-install/configure-host.sh
mcwienczek: I will try to add it

mcwienczek: done

amouat:
After learning my lesson the hard way, let me do a quick manual check before hitting merge...

quick-install/copy-certs.sh
```diff
@@ -3,16 +3,22 @@ set -eo pipefail
 unset CDPATH
 IFS=$'\n\t'
 
+namespace='kube-public'
+if [ ! -z "$1" ]
```
DavidZisky: doesn't necessarily need to be changed but "-n", in this case, would make a bit more sense

amouat: True. I'll merge a PR if you open one :)

PR #44, opened 2020-02-07
Update copy-certs.sh

DavidZisky:
Small redesign of if statement


amouat:
Travis failed, but it's due to a network problem in the CI, nothing to do with the code.

PR #45, opened 2020-02-10
WIP - sorting and limiting list_tag results

mcwienczek:
WIP still , but please review already


lib/protobuf/src/server.proto
```diff
@@ -81,6 +81,8 @@ message CatalogRequest {
 
 message CatalogEntry {
   string repo_name = 1;
+  int32 limit = 2;
+  string last_tag = 3;
```
amouat: I'm a bit surprised to see this in the return type. I think it should be in the request, but not the return?

lib/server/src/server.rs
```diff
@@ -672,20 +675,26 @@ impl Registry for TrowServer {
...
+                let it = paths.iter().take(limit);
+
+                for entry in it {
+                    let en_path = entry.path();
+                    if en_path.is_file() {
+                        if let Some(tag_str) = en_path.file_name() {
+                            let tag = Tag {
+                                tag: tag_str.to_string_lossy().to_string(),
+                            };
+                            tx.send(Ok(tag)).await.expect("Error streaming tags");
```
amouat: I'm sorry, I've wrecked this with the refactoring on master.

src/routes.rs
```diff
@@ -666,14 +666,21 @@ fn get_catalog(
...
     ci: rocket::State<ClientInterface>,
     repo_name: String,
+    limit: Option<i32>
 ) -> Result<TagList, Error> {
     let rn = RepoName(repo_name);
-    let tags = ci.list_tags(&rn);
+    let mut lim = std::i32::MAX;
+    match limit {
+        Some(i) => lim = i,
+        None => {}
```
amouat: I guess you could do
```
let limit = match limit {
  Some(l) => l,
  None => std::i32::MAX
}
```
Also std::i32::MAX seems quite a lot of results? Should we have a lower default value?

src/routes.rs
```diff
@@ -687,7 +694,7 @@ fn list_tags_2level(
     user: String,
     repo: String,
```
amouat: Should be `   limit: Option<i32>` here as well, right?

src/routes.rs
```diff
@@ -698,7 +705,7 @@ fn list_tags_3level(
...
```
amouat: Again    `limit: Option<i32>`

PR #48, opened 2020-02-11
Pin Dockerfiles and Travis to working version of rust.

amouat:
Need nightly with rustfmt.


mcwienczek:
What about ARM dockerfiles?

PR #53, opened 2020-02-14
Made error codes and returnning error messages conformant to Distribu…

### LINKED ISSUE #52: Make error codes conformant with Distribution Spec
Trow needs to be conformant with error format of Distribution Spec:
https://github.com/opencontainers/distribution-spec/blob/master/spec.md#errors-2

mcwienczek:
…tion Spec https://github.com/opencontainers/distribution-spec/blob/master/spec.md\#errors-2
Fixes #52

### DIFF
--- a/src/response/errors.rs
+++ b/src/response/errors.rs
@@ -49,28 +49,31 @@ struct ErrorMsg {
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match *self {
-            Error::Unsupported => write!(f, "Unsupported Operation"),
-            Error::Unauthorized => write!(f, "Authorization required"),
-            Error::BlobUnknown => write!(f, "Blob Unknown"),
+            Error::Unsupported => format_error_json(f, "UNSUPPORTED".to_string(), "Unsupported".to_string(), json!({})),
+            Error::Unauthorized => format_error_json(f, "UNAUTHORIZED".to_string(), "Authorization required".to_string(), json!({})),
+            Error::BlobUnknown => format_error_json(f, "BLOB_UNKNOWN".to_string(), "Blob Unknown".to_string(), json!({})),
             Error::BlobUploadUnknown => write!(f, "Blob Upload Unknown"),
-            Error::InternalError => write!(f, "Internal Error"),
-            Error::DigestInvalid => write!(f, "Provided digest did not match uploaded content"),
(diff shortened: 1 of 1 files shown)


src/response/errors.rs
```diff
@@ -49,30 +49,33 @@ struct ErrorMsg {
...
+            //TODO: INTERNAL_ERROR code is not in the distribution spec
+            Error::InternalError => format_error_json(f, "INTERNAL_ERROR".to_string(), "Internal Server Error".to_string(), json!({})),
+            Error::DigestInvalid => format_error_json(f, "DIGEST_INVALID".to_string(), "Provided digest did not match uploaded content".to_string(), json!({})),
+            Error::ManifestInvalid => format_error_json(f, "MANIFEST_INVALID".to_string(), "manifest invalid".to_string(), json!({})),
+            Error::ManifestUnknown(ref tag) => format_error_json(f, "MANIFEST_UNKNOWN".to_string(), "manifest unknown".to_string(), json!({ "Tag": tag }))
         }
     }
 }
 
+fn format_error_json(f: &mut fmt::Formatter, code: String, message: String, detail: serde_json::Value) -> fmt::Result {
```
amouat: This should take &str's I think

src/response/errors.rs
```diff
@@ -49,30 +49,33 @@ struct ErrorMsg {
...
+    let emsg = ErrorMsg {
+        code: code,
+        message: message,
+        detail: detail,
```
amouat: I wonder if we should change the detail to be `Option<String>`.

mcwienczek: I didn't do it just because I didn't know how the None would be translated to json. I can check it out.

mcwienczek: I wouldn't do that because then we are pushing responsiblity to generate proper json on the caller of the function.

mcwienczek: ideally, we could pass arbitrary struct here with details and that would be translated to json string, but I don't know how can I do that.

amouat: I'm not sure what you mean. I'm pretty sure it will all just work if you just change:
```
struct ErrorMsg {
    code: String,
    message: String,
    detail: Value,
}
```
to:

```
struct ErrorMsg {
    code: String,
    message: String,
    detail: Option<Value>,
}
```
Then if there's no details, you can just pass None. I think serde will still happily deal with converting to Json.

mcwienczek: please take a look now. It that, what you meant?

PR #54, opened 2020-02-14
Add support for doing uploads in POST.

amouat:


src/routes.rs
```diff
@@ -454,41 +457,84 @@ fn post_blob_upload(
...
+                        }
+                        Err(_) => Err(Error::InternalError),
+                    }
+                }
+                Err(_) => {
+                    // TODO: this conflates rpc errors with uuid not existing
+                    // TODO: pipe breaks if we don't accept the whole file
+                    // Possibly makes us prone to DOS attack?
+                    warn!("Uuid {} does not exist, piping to /dev/null", &up_info.uuid());
+                    let _ = data.stream_to_file("/dev/null");
```
mcwienczek: can't we just break the connection instead?

amouat: Not sure, I've not really looked into this tbh

amouat: I think this is best left for another PR - do you want to investigate?

mcwienczek: I will create an issue for it, meanwhile I will approve

src/routes.rs
```diff
@@ -454,41 +457,84 @@ fn post_blob_upload(
...
+                            rt.block_on(r).map_err(|_| Error::InternalError).map(|au| Upload::Accepted(au))
+                        }
+                        Err(_) => Err(Error::InternalError),
+                    }
+                }
+                Err(_) => {
+                    // TODO: this conflates rpc errors with uuid not existing
+                    // TODO: pipe breaks if we don't accept the whole file
+                    // Possibly makes us prone to DOS attack?
+                    warn!("Uuid {} does not exist, piping to /dev/null", &up_info.uuid());
```
mcwienczek: #55

PR #57, opened 2020-02-17
Drop connection rather than stream to /dev/null.

### LINKED ISSUE #55: Doing uploads in POST waits for invalid payload to stream
Doing uploads in POST streams all the data of an invalid payload to /dev/null. This can still strain the resources of trow instance. We should be breaking the connection once we already see that the payload is invalid.

See this line for more info: https://github.com/ContainerSolutions/trow/pull/54/files#diff-e12f07ef6d281c675e309912e464c925R494

amouat:
Closes #55.

### DIFF
--- a/src/routes.rs
+++ b/src/routes.rs
@@ -279,10 +279,7 @@ fn put_blob(
         }
         Err(_) => {
             // TODO: this conflates rpc errors with uuid not existing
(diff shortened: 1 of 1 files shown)


mcwienczek:
LGTM

PR #60, opened 2020-02-28
Typo error in file - copy-certs-tmpl

Pradyumnakashyap:
Typo error in the file (copy-certs-tmpl.YAML) causing the install script to fail. fixed the typo and now installation is successfull.


amouat:
Looks good. Thanks.

PR #97, opened 2020-04-24
First draft of contributing guidelines.

amouat:


amouat:
@cewood do you have time to take a look?

amouat:
Actually, something is better than nothing, so I've just merged this. Please let me know if you have any feedback @cewood

cewood:
Sorry, had a full schedule today. I'll take a look and ping you/or raise a PR if I have any worthy suggestions/feedback.

CONTRIBUTING.md
```diff
@@ -0,0 +1,63 @@
...
+Once you've got the go ahead to work on a feature, you can start work. Feel free to communicate with
+team via updates on the issue tracker and ask for feedback, pointers etc. Once you're happy with
+your code, go ahead and open a Pull Request.
+
+## Pull Request Process
+
+Most of the code is written in Rust and uses the standard style guidelines - please run your code
+through `rustfmt` prior to opening a PR.
+
+On opening a PR, a GitHub action will execute the test suite against the new code. All code is
```
cewood, line 48: I believe this isn't the case, I also thought this was the case and learned today that actions can't be run on forks, from what I understand :grimacing:

CONTRIBUTING.md
```diff
@@ -0,0 +1,63 @@
...
+discussing it first to avoid wasting effort. We do commit to listening to all proposals and will do
+our best to work something out!
+
+Once you've got the go ahead to work on a feature, you can start work. Feel free to communicate with
+team via updates on the issue tracker and ask for feedback, pointers etc. Once you're happy with
+your code, go ahead and open a Pull Request.
+
+## Pull Request Process
+
+Most of the code is written in Rust and uses the standard style guidelines - please run your code
```
cewood, line 45: My suggestion would be to add a Makefile, and or a Dockerfile to test the project in a repeatable way. Since you're using latest, this could vary by person depending on when they last updated. Pinning this to a version, in a Dockerfile, and having a not so complex Makefile to drive executing that locally, could be useful. I could contribute this if you're interested.

DEVELOPING.md
```diff
@@ -0,0 +1,12 @@
+# Developing Trow
+
+The easiest way to build and test Trow is by using the [Dockerfiles](./docker/README.md).
+
+Trow is written in [Rust](https://www.rust-lang.org/) and can be compiled, run and tested with the
```
cewood, line 5: Similar to above comments, I think a Hermetic local environment, for test/build in Docker could be useful.

cewood:
Late to the party, but looks good. Left a few comments, let me know if you have any thoughts on this.

CONTRIBUTING.md
amouat: What isn't the case? The tests definitely run on PR, I've been using that for a while.

amouat: If you have your own fork it won't run; it's just PRs against the trow repo.

CONTRIBUTING.md
amouat: I loathe Make with a passion :)

There are Dockerfiles, including a Dockerfile that runs the tests. I'll try to be a bit clearer in the text.

cewood: Oh Adrian, you were off to such a great start, and then... :grimacing: :laughing: 

Come to the dark side, Make can be your friend :wink:

amouat: I should mention that it's difficult to pin to a version of nightly - eventually you end with a dependency that doesn't work. So what we try to say is "the latest nightly that supports rustfmt and bunch of other components". It's a mess. The solution is moving to stable rather than pinning.

amouat: I wouldn't be against Make if it did something useful for us, but at the moment it would just be a complicated way to call single line commands.

Issue #67, opened 2020-03-11
Add readiness/liveness probes


amouat:
We should have readiness/liveness probes that work with kubernetes.

A basic endpoint is fine to begin with, more complicated metrics may be possible later.

amouat:
Closed by #116

PR #116, opened 2020-05-25
Feature/67 readiness and liveliness

iamcaleberic:
#67 
- Add /healthz route 
- Add readiness and liveliness pods to manifest


quick-install/trow-gke.yaml
```diff
@@ -100,6 +100,22 @@ spec:
           name: cert-vol
         - mountPath: /data
           name: data-vol
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            port: 8443
+            scheme: HTTPS
+            httpHeaders:
```
amouat, line 108: What are the headers for?

quick-install/trow.yaml
```diff
@@ -101,6 +101,25 @@ spec:
...
+        readinessProbe:
+          httpGet:
+            path: /healthz
+            port: 8443
+            scheme: HTTPS
+            httpHeaders:
+            - name: READINESS
+              value: "True"
+
+          initialDelaySeconds: 3
```
amouat, line 121: How did you come up with the numbers? Unless there's a reason, I think we should leave it as the default.

src/routes.rs
```diff
@@ -105,6 +106,15 @@ fn get_homepage<'a>() -> HTML<'a> {
     HTML(ROOT_RESPONSE)
 }
 
+/*
+* Trow health endpoint
+* GET /healthz
+ */
+#[get("/healthz")]
+fn get_healthz() -> JsonValue {
```
amouat, line 114: I think I'd prefer just `healthz` as the function name.

I added get/put etc to the OCI endpoints as it makes it easier to go between the spec and the code. This does mean `get_homepage` and `get_v2root` are wrong as well, but we can worry about that later.

quick-install/trow-gke.yaml
iamcaleberic: I added them incase you would need to tell apart livelivess calls

iamcaleberic:
> Looks a reasonable first version! I added some minor comments to the code.
> 
> Can we also add a simple test that just hits the healthz endpoint and verifies that it returns a 200?
> 
> I'm not sure about having readiness and liveness using the same endpoint - they're not really the same thing. I'm thinking we can probably get rid of the readiness check for the moment - trow should start within a second or so. WDYT?

I did not know how we were defining a readiness check for trow so it was more of a placeholder

src/routes.rs
iamcaleberic: Ok will update this

quick-install/trow.yaml
iamcaleberic: Will roll this back as well as the headers

src/routes.rs
```diff
@@ -105,6 +106,15 @@ fn get_homepage<'a>() -> HTML<'a> {
...
+    json!({ "status": "ok" })
```
Spazzy757, line 115: Just a comment on this Health endpoint, is there not more that can be tested? I.e that trow can write to the directory it needs to in order to store images?

Spazzy757: I know first version needs to be simple, so just a thought

amouat: I'm actually pretty happy with this as a liveness check - it verifies that the web server is up and responding. 

We should probably check that the backend gRPC server can be reached as well, but I'm also happy to leave that for a later PR.

I guess the write test should be in a readiness check?

iamcaleberic: Let me check on this

Issue #107, opened 2020-05-06
Change Trow UID


amouat:
The Trow UID is currently 999. This wasn't a good choice, as it maps onto existing users on some systems (it's systemd-coredump on my Ubuntu laptop). It would be better to use a high numbered user that doens't map onto anything.

Unfortunately this requires co-ordinating a change with the Docker images and the standard install YAML. If the user in the image changes without the YAML being updated, Trow won't run. For that reason, I'd like to make this change at the time of the next release (0.3), rather than immediately.

PR #129, opened 2020-06-08
Change uid from 999 to 333333.

amouat:
This is for security reasons - 999 typically maps to a system user on
the host, which gives attackers privileges in the case of a container
breakout. 333333 is unlikely to map onto anything, but respects the
/etc/user.defs convention of being in the range 10000-60000.


amouat:
Closes #107.

amouat:
Merging this. Release to follow later today.

Issue #126, opened 2020-05-29
Create Metrics Endpoint


amouat:
Should be compatible with prometheus.

Report on things like:
 - disk space
 - number of blobs, manifests, tags stored
 - number of requests served

amouat:
@iamcaleberic I think you've started work on this? Please link any PR to this issue.

PR #132, opened 2020-07-16
Feature/126 metrics

iamcaleberic:
#126 

- [x] Disk 

- [x]  Number of requests?
     - [x]  Blobs 
     - [x]  Manifests

- [ ] Number stored:
     - [ ] Manifests
     - [ ] Blobs
     - [ ] Tags


src/types.rs
```diff
@@ -339,3 +339,10 @@ pub struct ReadinessResponse {
     pub message: String,
     pub is_ready: bool
 }
+
+#[derive(Clone, Debug, Serialize, Deserialize ,PartialEq)]
+pub struct MetricsResponse {
```
amouat, line 344: I'm not sure about this.

I think it should probably only have the metrics string. The method then changes to return a `Result<MetricsResponse, Error>`. Or did that cause problems with setting up the routing?

tests/registry_interface.rs
```diff
@@ -474,6 +520,10 @@ mod interface_tests {
         tl4.insert("tag".to_string());
         tl4.insert("three".to_string());
         println!("Running check_tag_list_n_last 4");
-        check_tag_list_n_last(&client, 2, "latest", &tl4)
+        check_tag_list_n_last(&client, 2, "latest", &tl4);
+        
+        get_readiness(&client);
+        get_health(&client);
+        get_metrics(&client);
```
amouat, line 527: Could we add a test at the start that gets the diskspace metric, then at the end run it again and test that the value has changed (diskspace has been consumed)?

Also, add `println!` statements to help debugging. These tests should really be in separate test methods, but that causes problems with need to stop and restart and the registry, hence the `test_runner`. One of the downsides is that there's less information on what test is running, so the printlns help.

lib/server/src/statics.rs
```diff
@@ -0,0 +1,32 @@
+use lazy_static::lazy_static;
```
amouat, line 1: Let's rename this file. I didn't mean for it to just be statics, but rather all the metrics stuff that can be pulled out of `server.rs`. I guess just `metrics.rs` unless you have a better suggestion?

lib/server/src/server.rs
```diff
@@ -859,4 +879,38 @@ impl Registry for TrowServer {
         };
         Ok(Response::new(reply))
     }
-}
+
+    async fn get_metrics(
```
amouat, line 883: I'd say move this into `metrics.rs`, but I think that won't work as this is inside the `impl`.

lib/server/src/server.rs
```diff
@@ -580,6 +597,8 @@ impl Registry for TrowServer {
         //Don't actually need to verify here; could set to false
 
         let mr = req.into_inner();
+        // Increase manifest request metrics counter
```
amouat, line 600: Not sure this comment adds much.

lib/server/src/server.rs
```diff
@@ -25,6 +26,10 @@ use self::trow_server::*;
 use crate::server::trow_server::registry_server::Registry;
 
 
+use prometheus::{TextEncoder, Encoder};
+
+use crate::statics::{FREE_SPACE, AVAILABLE_SPACE, TOTAL_SPACE, TOTAL_MANIFEST_REQUESTS, TOTAL_BLOB_REQUESTS};
```
amouat, line 31: as below, rename statics to metrics. But what about changing this to `use crate::metrics;` and not importing the individual metrics? Then below you would write `metrics::FREE_SPACE.inc()` etc, which I think makes it a bit clearer what is going on.

lib/server/src/server.rs
```diff
@@ -203,6 +208,17 @@ fn get_digest_from_manifest_path<P: AsRef<Path>>(path: P) -> Result<String, Erro
         .to_string())
 }
 
+// Query disk metrics
+fn query_disk_metrics(path: &PathBuf) {
```
amouat, line 212: move this to `metrics.rs`

lib/server/src/server.rs
iamcaleberic: Refactored to move some of the logic away.

tests/registry_interface.rs
iamcaleberic: Added the print line. Not sure about the disk differences test because disk test though.

lib/server/src/metrics.rs
```diff
@@ -0,0 +1,66 @@
...
+    // * disk
+    // * total manifest requests
+    // * total blob requests
+
+    let metric_families = prometheus::gather();
+    let mut buffer = vec![];
+
+    encoder.encode(&metric_families, &mut buffer)?;
+
+    Ok(String::from_utf8(buffer).unwrap())
```
amouat, line 65: You return a result anyway, so why not use ? instead of unwrap?

tests/registry_interface.rs
```diff
@@ -474,6 +520,15 @@ mod interface_tests {
...
-        check_tag_list_n_last(&client, 2, "latest", &tl4)
+        check_tag_list_n_last(&client, 2, "latest", &tl4);
+        
+        println!("Running get_readiness");
+        get_readiness(&client);
+
+        println!("Running get_health");
+        get_health(&client);
+
+        println!("Running get_metrics");
```
amouat, line 531: can we add a test to make sure that the manifest and blob request numbers go up by the expected numbers? Just scrape the endpoint and save the numbers, then run again after getting a manifest .

lib/server/src/metrics.rs
iamcaleberic: True

amouat (changes requested):

amouat (approved):

amouat:
Sorry, there's a problem with `Cargo.lock` that needs fixed now. I think you just need to rebase and maybe run `cargo update`

PR #133, opened 2020-07-31
Added Basic User Setup

Spazzy757:
# Changelog

- Added User Struct
- Added Create Trait
- Added Authorize Trait
- Added SQlite Storage
- Added a bunch of test


amouat:
Does your setup run `rustfmt` automatically? I should set something up so that everything goes through `rustfmt`.

src/users.rs
```diff
@@ -0,0 +1,243 @@
...
+}
+
+// Constants
+const CREDENTIAL_LEN: usize = digest::SHA512_OUTPUT_LEN;
+const N_ITER: u32 = 100_000;
+
+// Error Used for User related Functions
+// Implements the traits of Error
+#[derive(Debug)]
+struct UserError(String);
```
amouat, line 25: I'm not convinced we need this, we can probably just return the existing errors. I've been using the `failure` crate, which we can continue to use for the minute, but there are better options now.

src/users.rs
```diff
@@ -0,0 +1,243 @@
...
+    pbkdf2::verify(
+        &digest::SHA512,
+        N_ITER,
+        &salt,
+        password.as_bytes(),
+        &pbkdf2_hash,
+    )
+}
+
+// creates a sqlite DB if it does not exist and initializes the user table
```
amouat, line 69: Creates the table, but not a whole DB, right?

src/users.rs
```diff
@@ -0,0 +1,243 @@
...
+// returns the connection
+fn connection() -> Result<rusqlite::Connection, Box<dyn Error>> {
+    let conn = Connection::open("sqlite.db").expect("db conn fail");
```
amouat, line 72: This should return an error, not use expect. In general we don't want to use `unwrap`/`expect` as they cause an immediate panic. There are times this is the right thing to do, but in this case I think we should return an error and let the caller decide.

amouat: You should be able to use `failure::error` in the return type and then just use the `?` operator here.

src/users.rs
```diff
@@ -0,0 +1,243 @@
...
+             hash text not null,
+             active integer default 1 not null
+         );",
+        NO_PARAMS,
+    ) {
+        Ok(_) => {
+            return Ok(conn);
+        }
+        Err(_) => {
+            return Err(Box::new(UserError("Error Creating table".into())));
```
amouat, line 89: Is there a reason (e.g. security) you don't want to return the original error? Then you can do:

```
conn.execute(...)?;
Ok(conn)
```

src/users.rs
```diff
@@ -0,0 +1,243 @@
...
+    // trait used to create a user
+    pub fn create(&mut self, password: String) -> Result<(), Box<dyn Error>> {
+        // Generates a salt for the user
+        let salt = get_salt();
+        self.salt = HEXUPPER.encode(&salt);
+        // Generates a hash from the salt and password
+        let hash = get_hash_from_password(password, salt);
+        self.hash = HEXUPPER.encode(&hash);
+        // Connect to the db
+        match connection() {
```
amouat, line 104: Same as above regarding error handling - can we just return the original errors?

src/users.rs
```diff
@@ -0,0 +1,243 @@
...
+    }
+    // Trait to Authorize the User
+    pub fn authorize(&mut self, password: String) -> Result<(), Box<dyn Error>> {
+        // Connect to DB
+        match connection() {
+            Ok(conn) => {
+                // Prepare Select statement
+                let mut stmt = conn
+                    .prepare("SELECT id, name, salt, hash, active FROM users WHERE name = :name LIMIT 1;")
+                    .expect("");
```
amouat, line 134: You could leave this as it should never error, but I think `?` would be cleaner as other stuff can return an error anyway. Also you may as well call `unwrap` if you're not using the string.

amouat: Same goes for the following `expects`. All of these will crash Trow if they ever occur.

src/users.rs
```diff
@@ -0,0 +1,243 @@
...
+                // Run the Query Passing in the user name
+                let mut rows = stmt
+                    .query_named(named_params! { ":name": self.name })
+                    .expect("");
+                // Iterate over results (Should only ever return one)
+                while let Some(row) = rows.next().expect("") {
```
amouat, line 140: This one worries me - if I give the code a non-existent user, does it panic?

src/users.rs
```diff
@@ -0,0 +1,243 @@
...
+            "Expected password to get verified"
+        );
+        assert!(
+            !verify_incorrect_password.is_ok(),
+            "Expected password to fail verification"
+        )
+    }
+
+    #[test]
+    fn test_create_and_get() {
```
amouat, line 214: Can we add some tests for failing cases? e.g. trying to get a user that doesn't exist?

Spazzy757: Yes sir :)

src/users.rs
Spazzy757: Will remove

src/users.rs
Spazzy757: Sqlite will create a DB file if it does not exist when you try open it, this is default sqlite behaviour

src/users.rs
Spazzy757: okay awesome, I was a bit confused with the error handeling

src/users.rs
Spazzy757: My thinking was that there are a bunch of different errors that can be returned just normalize them all to one type of error, but I will do your suggestion above

src/users.rs
Spazzy757: Will do this :)

Spazzy757:
> Does your setup run `rustfmt` automatically? I should set something up so that everything goes through `rustfmt`.

The [Spacemacs Rust](https://develop.spacemacs.org/layers/+lang/rust/README.html) layer is pretty awesome

src/users.rs
```diff
@@ -0,0 +1,254 @@
...
+             active integer default 1 not null
+         );",
+        NO_PARAMS,
+    )?;
+    Ok(conn)
+}
+
+impl User {
+    // trait used to create a user
+    pub fn create(&mut self, password: String) -> Result<(), failure::Error> {
```
amouat, line 68: I'm thinking this should maybe change to `new(name: String, password: String) -> Result<User, Error> { ...`

Then take the `pub` off the struct members. I don't _think_ the caller is interested in setting anything except name and password, right?

Spazzy757: at this stage I dont beleive so, I will update this ASAP

Spazzy757: Done, @amouat I updated all the traits to follow this as I felt it was a better pattern

amouat:
So to compile Trow we now need the sqlite lib, right?

I'm wondering if we should put this behind a feature. It's just means adding something like:
```
#[cfg(feature = "sqlite")]
```

before the `mod users` declaration. Then if someone doesn't want to pull in sqlite or their platform doesn't have the library, they can still build it. WDYT? It will probably get a bit more complicated as we integrate the code.

Spazzy757:
> So to compile Trow we now need the sqlite lib, right?
> 
> I'm wondering if we should put this behind a feature. It's just means adding something like:
> 
> ```
> #[cfg(feature = "sqlite")]
> ```
> 
> before the `mod users` declaration. Then if someone doesn't want to pull in sqlite or their platform doesn't have the library, they can still build it. WDYT? It will probably get a bit more complicated as we integrate the code.

I'm not 100% sure about this as if it adds complications later on is it worth it to save a mb or two 🤷‍♂️  but if you want to add it i can

amouat:
> > So to compile Trow we now need the sqlite lib, right?
> > I'm wondering if we should put this behind a feature. It's just means adding something like:
> > ```
> > #[cfg(feature = "sqlite")]
> > ```
> > 
> > 
> > before the `mod users` declaration. Then if someone doesn't want to pull in sqlite or their platform doesn't have the library, they can still build it. WDYT? It will probably get a bit more complicated as we integrate the code.
> 
> I'm not 100% sure about this as if it adds complications later on is it worth it to save a mb or two man_shrugging but if you want to add it i can

I'm not sure either.

It's not the size so much, it's that I have to add the sqlite libraries on my host (`sudo apt install libsqlite3-dev`). The apt package is only 2,372 kB uncompressed though.

amouat:
For the moment, let's keep it behind a flag, otherwise people will need to install the sqlite libs on their host, even though they're not currently used. 

They are pretty popular and small though, so we can probably get rid of it once the feature is finished.

Spazzy757:
> For the moment, let's keep it behind a flag, otherwise people will need to install the sqlite libs on their host, even though they're not currently used.
> 
> They are pretty popular and small though, so we can get rid of it once the feature is finished (and update the build instructions to mention the dependency here https://github.com/ContainerSolutions/trow/blob/master/DEVELOPING.md).
> 
> I know the Docker build works ok with it, but it seems `rust:latest` has the sqlite library already for some reason.

@amouat Added you just need to check if its correct, also pulled in the dependency updates from master

PR #137, opened 2020-08-12
Deps update

silenteh:
Working in progress to update Trow dependencies


docker/Dockerfile.test
```diff
@@ -2,7 +2,7 @@
 FROM rust:latest as builder
 
 RUN rustup default nightly-2020-02-18 && rustup component add rustfmt
-RUN apt-get update && apt-get install -y unzip
+RUN apt-get update && apt-get install -y unzip libssl-dev
```
amouat, line 5: What's pulling libssl in?

src/lib.rs
```diff
@@ -217,13 +219,14 @@ impl TrowBuilder {
...
             .keep_alive(0) // Needed to mitigate #24. See above.
-            .secret_key(skey)
+            .secret_key(secret_key)
             .workers(256);
 
         if let Some(ref tls) = self.config.tls {
             if !(Path::new(&tls.cert_file).is_file() && Path::new(&tls.key_file).is_file()) {
                 return  Err(format_err!("Trow requires a TLS certificate and key, but failed to find them. \nExpected to find TLS certificate at {} and key at {}", tls.cert_file, tls.key_file));
             }
+            println!("{} - {}", tls.cert_file.clone(), tls.key_file.clone());
```
amouat, line 229: Did you mean to delete this?

tests/authentication.rs
```diff
@@ -35,7 +35,7 @@ mod authentication_tests {
     /// Call out to cargo to start trow.
     /// Seriously considering moving to docker run.
 
-    fn start_trow() -> TrowInstance {
+    async fn start_trow() -> TrowInstance {
```
amouat, line 38: I'm not sure it's worth moving all the test code to async hyper. Isn't making it more complicated?

lib/server/src/digest.rs
```diff
@@ -0,0 +1,52 @@
+use failure::{self, Error};
+use std::io::Read;
+
+// Crypto and crypto related imports
+use sha2::{Sha256, Digest};
+
+// Buffer size for SHA2 hashing
+const BUFFER_SIZE: usize = 1024;
+
+fn digest<D: Digest + Default, R: Read>(reader: &mut R) -> Result<String, Error> {
```
amouat, line 10: thanks for pulling this out to a separate file, it helps.

docker/Dockerfile.test
silenteh: I was trying to figure out why running tests with TLS did not work. I thought the Docker crash was caused by a missing dependency. It turned out it was a crash caused by the RAM limit of my Docker configuration. I removed it.

src/lib.rs
silenteh: Yep same as above. Was debugging the TLS part. It's a left over. Good catch. Removed it.

tests/authentication.rs
silenteh: We are using a more recent version of `reqwest` which is async by default. So whenever we use reqwest now we need to make the function async.  
`start_trow` is using `reqwest` to verify that Trow has started. This is why I have to use `async` everywhere.

lib/server/src/digest.rs
silenteh: No problem.  
Since I was moving away from rust-crypto crate, I thought it was a good idea to adjust a bit the digest part and write tests to make sure the digest was calculated correctly.

tests/authentication.rs
amouat: I think you can still turn it on by enabling the "blocking" feature. Maybe it's not worth it though.

PR #138, opened 2020-08-15
Allow configuring Trow's validation through Helm

paxswill:
I think this is the right branch, it seems this branch had more recent work on the Helm chart.

This PR adds the ability to modify the default validation configuration options through some new Helm chart values. Because I put these values under it, I also renamed the `trow.validatingWebhooks` value map to `trow.validation`.

I'm still getting a good handle on Kubernetes, so any guidance is greatly appreciated. I'm also not entirely sure I picked the right values for `admissionReviewVersions` and `sideEffects` for the validating webhooks, so some extra review on that would be appreciated.

### DIFF
--- a/charts/trow/templates/statefulset.yaml
+++ b/charts/trow/templates/statefulset.yaml
@@ -32,6 +32,30 @@ spec:
           - {{ .Values.trow.user | quote }} 
           - "--password-file" 
           - "/etc/trow/pass"
+        {{- with .Values.trow.validation }}
+          {{- if .allowDocker }}
+          - "--allow-docker-official"
+          {{- end }}
+          {{- if not .allowKubernetes }}
+          - "--deny-k8s-images"
+          {{- end }}
+          {{- if .allowPrefixes }}
+          - "--allow-prefixes"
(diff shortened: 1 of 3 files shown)


Spazzy757:
Hi @paxswill thank you for the contribution! Just a quick note, the branch `feature-helm-chart` was merged by [PR#122](https://github.com/ContainerSolutions/trow/pull/122) the commits where squashed into one https://github.com/ContainerSolutions/trow/commit/ad4964b2ee9fe1e74f8be22807b96c26600300eb. It is scheduled for deletion, could you open your PR against Master instead?

paxswill:
Sorry, didn't realize the Helm commits had been squashed already. I've rebased my branch onto master and changed this PR to master as well.

Spazzy757:
@paxswill Thanks! I just need to test this

Spazzy757 (approved):
All tested and seems to work thanks!

amouat:
Great, I'll click the big shiny merge button then :)

@paxswill thanks a lot for the contribution!

PR #179, opened 2020-09-29
Update issue templates

Spazzy757:
Added a feature request template to the issues templates


Spazzy757:
@amouat This is just so that we don't only have bug templates but also feature request templates (I wanted to do a feature request and didn't have the right template)

amouat (approved):

Issue #171, opened 2020-09-18
Helm chart: allow configuration with no user/pass


amouat:
The Helm chart currently asks for a user and password, but it should also be possible to configure with no user/password set at all.

alex-glv:
@amouat I can't reproduce this. Could you post exact steps?

amouat:
You can install from the Helm chart without providing username/pass? It didn't look possible to me - maybe it's just a docs issue.

Spazzy757:
@amouat Question on this, would you want the default for the helm chart to be no auth?

amouat:
Yes, I *think* so. But either is fine as long as it's easy to change.

PR #180, opened 2020-09-30
Feature helm Ci

Spazzy757:
## Problem

Fixes this issue https://github.com/ContainerSolutions/trow/issues/171 

Also added CI to test the Helm Chart


amouat (approved):
LGTM!

Let's merge it and see what happens :)

amouat:
I think we should run on every branch so PRs get checked.

PR #196, opened 2020-10-13
Proper fix for flaky test

amouat:


amouat:
Maybe the correct fix is on the service rather than pod? I note it still didn't work straight-away

PR #193, opened 2020-10-09
SPIFFE Implementation PoC - needs crate patches

blaggacao:
As promised: https://github.com/ContainerSolutions/trow/issues/189#issuecomment-706323292 /cc @imacg 

### Try it out

**Note:** _this tooling trades ease of use (a.k.a. reproducability) for disk space. expect this closure of this nix-shell to be about 400-600MB._

&rarr; have a look at the [updated quick-install](https://github.com/ContainerSolutions/trow/blob/9d42e01225da27e85cca2f4fd6f4b57ba995400d/QUICK-INSTALL.md)

```console
$ curl -L https://nixos.org/nix/install | sh
$ cd .../trow
$ nix-shell
$ cd ./dev
$ menu # informative
...
$ just-menu # informative
...
$ dns-activate # see visuals below
...
$ ca-install # see visuals below
...
$ just quick-install # launches a 100% reproducible dev cluster based on k3d
...

$ # explore

$ just tear-down
...
$ dns-deactivate
...
$ ca-uninstall
...
```
<details>
<summary><strong>Some Visuals</strong></summary>

## switch on
image

## switch off
image

</details>

PR #205, opened 2020-10-21
Use named ports

blaggacao:
Named ports helps with better reasoning over the service's port
abstractions.

### DIFF
--- a/install/base/service.yaml
+++ b/install/base/service.yaml
@@ -7,7 +7,8 @@ spec:
     app: trow
   type: NodePort
   ports:
-  - protocol: TCP
+  - name: http-tcp
+    protocol: TCP
     port: 8000
-    targetPort: 8000
+    targetPort: trow-http-port
 
--- a/install/base/stateful-set.yaml
+++ b/install/base/stateful-set.yaml
@@ -25,7 +25,8 @@ spec:
           - "/etc/trow/pass"
         imagePullPolicy: Always
         ports:
-        - containerPort: 8000
+        - name: trow-http-port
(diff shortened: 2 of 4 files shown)


amouat:
Yes, definitely a good idea, but I'm a bit concerned with changing the service port to `80` from `8000`. I'm trying to figure out when that port is used and I'm not sure. It would be used for internal k8s communication, so may affect the health check for example.

blaggacao:
The service port 80 (or 443) is only used when the service is accessed directly. That only is the cade for a cluster wide ingress controller (since containerd/ etc) are not plugged into the k8s dns when pulling images. So any "in cluster" fetching would have to go through node port anyway (localhost:<nodeport>).

In order not to (implicitly) insinuate it's relevant we probably should leave it at 8000. I'll fixup.

amouat:
Yeah, it can't affect the health check as that must be implemented at the container level (I think even below the pod level).

It probably depends on the ingress controller - I think most of them try to avoid the overhead of going through the service and enumerate the pods directly (at least that's what the nginx ingress used to do).

blaggacao:
So I forced pushed and reverted back to 8000 to avoid any tacit [not-yet-vested](https://github.com/ContainerSolutions/trow/issues/211) implications.

amouat:
I'm about to merge this.

In general the installers need more testing with various platforms. We may be able to extend the existing Helm tests with microk8s to cover some cases, but others need tests with specific platforms e.g. GKE which is more difficult to automate.

amouat (approved):

PR #206, opened 2020-10-21
Add https port definition

blaggacao:
Atop #205 

That keeps downstream kustomization files a little cleaner.

Maybe there are ways to also add it to the two ingress controllers
(as convenience defaults) without breaking anything?

if this is at risk of causing any troubles, better let's drop it.

### DIFF
--- a/install/base/service.yaml
+++ b/install/base/service.yaml
@@ -11,4 +11,8 @@ spec:
     protocol: TCP
     port: 8000
     targetPort: trow-http-port
+  - name: https-tcp
+    protocol: TCP
+    port: 8443
+    targetPort: trow-https-port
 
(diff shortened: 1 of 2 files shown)


amouat:
Yes, this probably makes sense.

I've been meaning to go through the install and make it use TLS between ingress and Trow, which isn't the case by default atm.

blaggacao:
> I've been meaning to go through the install and make it use TLS between ingress and Trow, which isn't the case by default atm.

Wanting to use spiffe for that, I got stuck with current lack of ECDSA cert support.

blaggacao:
So I rebased (8000 / 8443) to not to induce any not-yet-vested implications. I'm happy to hold this in stock or close as for when you'd want to make progress on this topic.

amouat:
@blaggacao could you resolve the conflicts and I'll merge this?

blaggacao:
@amouat Done. Thank you!

amouat (approved):

PR #207, opened 2020-10-21
Don't use nodeport service for production dpeloyments

blaggacao:
In production, users should not be "nudged" into using node ports.
Rather a proper deployment would include a proper ingress controller.

_I might be missing something, but I'll put this on the table to at least discuss and gain clarity._

### DIFF
--- a/install/base/service.yaml
+++ b/install/base/service.yaml
@@ -5,7 +5,6 @@ metadata:
 spec:
   selector:
     app: trow
-  type: NodePort
   ports:
   - protocol: TCP
     port: 8000


amouat:
Yes, that's a good point. 

It's an escape hatch that allows Nodes to reach the registry at the address `localhost:32000` or whatever. But production installs should not be using it.

I need to test this and figure out when it's used and when it's not. This should perhaps be converted into an issue rather than a PR.

blaggacao:
In my most recebt tests, I fugured that (at least) containerd has no access to the k8s dns. That means that a mirror pointing to `trow-deploy.trow-dev.svc.cluster.local` cannot be resolved without further dns setup (and in the case cni network policies are inplace probably even more complex setup).

So I'm inclined to think that this suggestion is actually wrong. I'm closing and will add an issue for tracking this branch of investigation.

PR #208, opened 2020-10-21
Add deployment base variant

blaggacao:
Some people might indeed wish to deploy a simplified stateless trow
in production. Especially so, if trow is (will be) only used as a proxy.

I for once, would prefer to deploy a stateless trow for my use case,
since it saves me the hassle to deal with all the extra work required to
make pvc portable across machines. (like deploying on longhorn,
openEBS or similar).


awoimbee:
Stateless Trow would be nice, but this PR is way too old, the project changed quite a bit !

PR #209, opened 2020-10-21
Unecesary configuration

blaggacao:
From looking at:
https://github.com/ContainerSolutions/trow/blob/master/quick-install/validate-tmpl.yaml#L6

It appeared to me that this configuration is not necessary and the value can be fixed to `validator.trow.io`.

Changing it appeared to me a cosmetic thing (which should not induce users into thinking it might be actually relevant). Am I missing something?

If this is ok, I should probably update docs and polish this further.

### DIFF
--- a/install/base/kustomization.yaml
+++ b/install/base/kustomization.yaml
@@ -39,13 +39,3 @@ patchesJson6902:
 #        name: trow-set
 #        group: apps
 #        version: v1
-#
-#    - patch: |-
-#        - op: replace
-#          path: /webhooks/0/name
-#          value: newregistry.mydomain.com
-#      target:
-#        kind: ValidatingWebhookConfiguration
-#        name: trow-validator
-#        group: admissionregistration.k8s.io
-#        version: v1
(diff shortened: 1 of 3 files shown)


amouat:
No, this isn't correct I think. The validator runs as part of Trow and will typically be the same as the Trow address (but could be different). `validator.trow.io` is just the default example assuming a registry running at `trow.io`.

blaggacao:
> The name of the admission webhook. Name should be fully qualified, e.g., imagepolicy.kubernetes.io, where "imagepolicy" is the name of the webhook, and kubernetes.io is the name of the organization. Required.

https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#validatingwebhook-v1-admissionregistration-k8s-io

I still think your conclusion is not mandatory, that means `validator.trow.io` can be a valid fqdn for this webhook while still properly connecting to the service that implements the webhook.

Not meaning to pick on it, but on the other hand, loosening this constraint would make deployment a little "simpler" since the operator can choose to modify those values or not to. ;-)

amouat (approved):

PR #226, opened 2021-01-10
Updated the chart to support using the new 0.3.1-PROXY image to use trow as a proxy for docker hub

22ndtech-admin:
I updated the chart version and the appVersion. I also added values to the values.yaml and parameters to the statefulset.yaml  to accept them for passing the image tag and the --proxy-docker-hub arguments.

### DIFF
--- a/charts/trow/values.yaml
+++ b/charts/trow/values.yaml
@@ -21,7 +21,7 @@ trow:
     allowImages: []
     disallowLocalPrefixes: []
     disallowLocalImages: []
-    proxyDockerHub: true
+    proxyDockerHub: false
 
 imagePullSecrets: []
 nameOverride: ""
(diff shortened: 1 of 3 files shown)


amouat:
Sorry I missed this! This looks good - let me get is tested and get back to you.

charts/trow/values.yaml
```diff
@@ -20,6 +21,7 @@ trow:
     allowImages: []
     disallowLocalPrefixes: []
     disallowLocalImages: []
+    proxyDockerHub: true
```
amouat, line 24: I think it should probably be false by default

amouat (changes requested):

charts/trow/values.yaml
22ndtech-admin: Ok, I changed it to false

amouat (approved):

amouat:
Merged, apologies for the delay.

Many thanks for the PR!

PR #227, opened 2021-01-14
Replaces user and PAT with secret values for easier maintenance

alex-glv:
Replaces login to ghcr.io user with the value from secret, so it's easier to update. 
Renames the token to CR_PAT as per github how to (not very important, but makes it more consistent) - https://docs.github.com/en/free-pro-team@latest/packages/guides/pushing-and-pulling-docker-images#authenticating-to-github-container-registry

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -14,7 +14,7 @@ jobs:
     - uses: actions/checkout@v2
     - name: Login to package repos
       run: |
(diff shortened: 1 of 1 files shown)


amouat:
This is using the cs-trow user right?

amouat (approved):
Looks good! I assume it maps to the new cs-trow user.

amouat:
Merging as it's broken anyway...

PR #230, opened 2021-01-28
feature/gui

iamcaleberic:
- [x] -   Handle authentication flow
- [x] -   display the images the user has access to and various info/stats like annotations
- [ ] -   support various admin tasks such as deleting images or giving users access to repos etc
- [ ] -   display the results of security scans etc
- [ ] -   display the history of tags and images
- [ ] -   allowing searching of images


amouat:
This looks good! I'm actually wondering if we should put the GUI in a separate repository. What do you think?

iamcaleberic:
> This looks good! I'm actually wondering if we should put the GUI in a separate repository. What do you think?

I did not see this earlier,  we can have it a separate repo. I don't think it will be too inconvenient.

PR #231, opened 2021-02-08
Update rust version and libs.

mathijshoogland:


amouat:
This looks good. What was the seek_stream_len thing?

mathijshoogland:
When building the project with the latest rust nightly version project wasn't able to successfully build because apparently _seek_stream_len_ function was declared unstable. CLI gave feedback that, if we are running on nightly, to simply declare it at the crate root. (At the same time, seek_convenience apparently isn't unstable anymore). Since versions are pinned in docker this only is an issue when building the project locally (or in our CI/CD pipeline).

To see an example of the error you can have a look at the following build:
https://github.com/ContainerSolutions/trow/actions/runs/552531497

amouat:
Thanks, I ran it and went through the issues. I don't fully understand, but it's seems the unstable seek features were split up and one of them was stabilised (but not stream_len). I think this is correct

amouat (approved):
Good work, this all makes sense!

PR #236, opened 2021-02-27
Set content-length for manifest and blobs

amouat:


amouat:
Need to add a test; there seems to be an off-by-one issue possibly in a dependent library :(

amouat:
I'll probably re-do this in a new branch...

PR #240, opened 2021-03-18
Trait refactor

amouat:
Refactored client_interface to use traits.

Still work to be done:
 - routes should use trait not concreate type
 - more refactorings/clean ups on client interface


amouat:
Why do I always fuck up rebases? This time I didn't update master :facepalm:

amouat:
Fingers crossed that this is ok. :crossed_fingers:

PR #249, opened 2021-04-23
Reverted some changes of PR#237.

mathijshoogland:
So I think this was merged by accident, we don't want this, it pins version 0.3.1-proxy in our chart. 

This was probably submitted because our helm chart wasn't updated for some time.

### DIFF
--- a/charts/trow/templates/statefulset.yaml
+++ b/charts/trow/templates/statefulset.yaml
@@ -22,7 +22,7 @@ spec:
     {{- end }}
       containers:
       - name: trow-pod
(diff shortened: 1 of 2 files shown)


amouat (approved):

amouat:
Thanks, I've merged this. I really don't know the proper way to manage Helm charts, so I generally just merge what I'm giving here!

In the future I'll ask for your review on any Helm PRs if that's ok.

amouat:
@mathijshoogland can you confirm that the chart is working and is running the latest release?

mathijshoogland:
> working and is running the latest release?

Confirmed

PR #255, opened 2021-06-18
skip only blob which is already downloaded

sopak:
we need to continue as docker manifests may share blobs
for example they may have same FROM


amouat:
I think the failure must be related to out-of-date libs. Rerunning.

amouat (approved):

amouat:
Interesting - it still runs against the old base. I could ask you to rebase but it seems a bit unnecessary. Let me merge and hope for the best :)

PR #250, opened 2021-05-28
Implement kubernetes.io/dockerconfigjson as configuration for proxy any number of registries

sopak:
for every f/anyname/ try to do the proxy according configuration directory

Local directory with configuration for proxy any registry at f/${registryCfgDir}/<repo_name> to anyregistry.xyz/<repo_name>. Downloaded images will be cached.
Directory should contain sub directories which should correspond to ${registryCfgDir}.
Each of them contains file .dockerconfigjson which should be in format as of kubernetes.io/dockerconfigjson.
The intention is to directly support mounting k8s secrets (imagePullSecrets) in kubernetes and act as proxy according to their configuration.
As of now config dir is parsed each request, to be able  to reflect changes in secrets. It may change in future."

PR #256, opened 2021-06-18
Refactor proxy authentication mechanism

sopak:
simplified way howto authenticate to proxy hub and other registries in following changes

this is here to split PR  from https://github.com/ContainerSolutions/trow/pull/250 to multiple steps

please check it and I will fix  what I can fort  this part

### DIFF
--- a/lib/server/src/server.rs
+++ b/lib/server/src/server.rs
@@ -1,28 +1,38 @@
-use crate::manifest::{FromJson, Manifest, manifest_media_type};
-use chrono::prelude::*;
-use failure::{self, Error, Fail};
-use prost_types::Timestamp;
+use core::fmt::Display;
+use std::collections::HashMap;
 use std::collections::HashSet;
 use std::fmt;
 use std::fs::{self, DirEntry, File};
 use std::io;
 use std::io::{BufRead, BufReader, Read, Write};
 use std::path::{Path, PathBuf};
+use std::str;
 use std::sync::{Arc, RwLock};
+
+use chrono::prelude::*;
+use failure::{self, Error, Fail};
+use futures::TryFutureExt;
+use prost_types::Timestamp;
+use quoted_string::strip_dquotes;
+use reqwest::{self, header::{HeaderMap, HeaderValue}, StatusCode};
+use reqwest::header::ToStrError;
+use rustc_serialize::hex::ToHex;
+use rustc_serialize::json::ToJson;
+use serde_json::Value;
 use tokio::sync::mpsc;
 use tonic::{Request, Response, Status};
 use uuid::Uuid;
-use core::fmt::Display;
-use reqwest::{self, header::{HeaderMap, HeaderValue}};
+
+use crate::digest::sha256_tag_digest;
+use crate::manifest::{FromJson, Manifest, manifest_media_type};
+use crate::metrics;
+use crate::server::trow_server::registry_server::Registry;
+
+use self::trow_server::*;
+
 pub mod trow_server {
     include!("../../protobuf/out/trow.rs");
 }
-
-use self::trow_server::*;
-use crate::digest::sha256_tag_digest;
-use crate::server::trow_server::registry_server::Registry;
-
-use crate::metrics;
 
 static SUPPORTED_DIGESTS: [&str; 1] = ["sha256"];
 static MANIFESTS_DIR: &str = "manifests";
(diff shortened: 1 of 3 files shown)


amouat:
This looks good. The major thing that's missing is some tests. Do you have tests working at the minute if you run `cargo test`? If not, I'll help you set it up.

lib/server/src/server.rs
```diff
@@ -1,28 +1,38 @@
...
 use std::fs::{self, DirEntry, File};
 use std::io;
 use std::io::{BufRead, BufReader, Read, Write};
 use std::path::{Path, PathBuf};
+use std::str;
 use std::sync::{Arc, RwLock};
+
+use chrono::prelude::*;
+use failure::{self, Error, Fail};
+use futures::TryFutureExt;
```
amouat, line 14: Can you fix all the unused imports?

lib/server/src/server.rs
```diff
@@ -573,11 +605,14 @@ impl TrowServer {
...
             //Get auth token
 
-            let auth_token = if let Some(auth) = &proxy_auth {
-                self.get_auth_token(&cl, &proxy_image, &auth).await
-            } else {
-                None
+            let auth_token = match self.get_auth_token(&cl, &proxy_image, &proxy_auth).await {
+                Ok(a) => Some(a),
+                Err(e) => {
+                    error!("Can't get auth_token: {}", e);
```
amouat, line 611: Let's expand this a bit so it makes more sense to the user e.g. "Failed to get auth token for IMAGE with error ERR"

lib/server/src/server.rs
```diff
@@ -474,65 +478,93 @@ impl TrowServer {
...
+            .query(
+                &bearer_param_map
+            );
 
-            cl.post(&auth.endpoint)
-                .header("Content-Type", "application/x-www-form-urlencoded")
-                .body(auth_req).send().await
+        let reqBuilder = match auth {
+            Some(a) => {
+                if a.user.is_some() {
```
amouat, line 503: I think this can be refactored to get rid of the unwraps. Something like `if let Some(a_user) = &a.user {`

Then use `a_user` instead of `a.user`

lib/server/src/server.rs
```diff
@@ -474,65 +478,93 @@ impl TrowServer {
...
-                error!("Failed to find auth token in auth repsonse"); 
-                None
-            }
+        let auth_json_body = resp.json::<serde_json::Value>().await
+            .or_else(|e |Err(format_err!("Failed to deserialize auth response {}", e)))?;
+
+        let access_token= auth_json_body.get("access_token")
+            .ok_or(format_err!("Failed to find auth token in auth repsonse"))?;
+
+        debug!("access_token: {}",access_token);
```
amouat, line 526: Either remove this or add more context

lib/server/src/server.rs
```diff
@@ -474,65 +478,93 @@ impl TrowServer {
...
+
+        access_token.as_str()
+            .map(|s| s.to_string())
```
amouat, line 529: I don't think we need map here?

lib/server/src/server.rs
```diff
@@ -474,65 +478,93 @@ impl TrowServer {
...
+            .ok_or(format_err!("Access token string conversion failed"))
+    }
+
+    async fn get_www_authenticate_header(&self, cl: &reqwest::Client, image: &Image) -> Result<String, Error> {
+        let resp = cl.head(&image.get_manifest_url())
+            .headers(create_accept_header())
+            .send().await
+            .or_else(|e| Err(format_err!("Request for authenticate failed: {}", e)))?;
```
amouat, line 537: Not sure if the reqwest error will include it, but it would be nice to put the URL in the error message.

lib/server/src/server.rs
```diff
@@ -474,65 +478,93 @@ impl TrowServer {
...
+            return Err(format_err!("Request '{}' should fail with status unauthorized", &image.get_manifest_url()));
         }
+
+        let value = resp
+            .headers()
+            .get("www-authenticate")
+            .ok_or(format_err!("Request header www-authenticate for authenticate should exists"))?;
+
+        value.to_str()
+            .map(|s| String::from(s))
```
amouat, line 549: Again, I don't think we need map here, maybe I'm missing something

lib/server/src/server.rs
```diff
@@ -474,65 +478,93 @@ impl TrowServer {
...
+        bearer_param_map.remove("realm");
 
-            if let Some(user) = &auth.user {
-                info!("Making request to Docker Hub as {}", &user);
-                auth_req = format!("{}&username={}", auth_req, user);
-            }
-            if let Some(pass) = &auth.pass {
-                auth_req = format!("{}&password={}", auth_req, pass);
-            }
+        let reqBuilder = cl.get(realm.as_str())
```
amouat, line 496: please use rust naming i.e. req_builder

amouat (changes requested):
Can you clean this up per the comments? Also really need some tests.

amouat:
@sopak could you e-mail me at adrian.mouat@container-solutions.com? I'm keen to help get this merged and I'd like to understand your use case.

BTW I realised `fmt` wasn't running on the server code when looking at this, so I've just enabled it. Unfortunately that means there is a bunch of conflicting formatting changes on the main branch which will be a pain. It's probably worth rebasing and/or running `cargo fmt` yourself. Sorry for the hassle, but it will make merges easier in the long term.

amouat:
@sopak I've updated this and will merge in a new PR.

PR #262, opened 2021-06-24
feature/cors

iamcaleberic:
PR: 

- Allows Cross-Origin requests to support gui

Checklist
- [x] Default values for headers and methods
- [x] Reduce required cor flags to two
- [x] Add tests


src/routes/tags.rs
```diff
@@ -0,0 +1,29 @@
+use rocket_contrib::json::{Json, JsonValue};
+
+#[options("/v2/<repo_name>/tags/list")]
+pub fn options_tags(
+    repo_name: String,
+) -> Json<JsonValue> {
```
amouat, line 6: I think it should CORS requests should have an empty body, not {}

src/routes/mod.rs
```diff
@@ -114,3 +128,8 @@ fn no_auth(_req: &Request) -> Authenticate {
 fn login(auth_user: ValidBasicToken, tc: State<TrowConfig>) -> Result<TrowToken, Error> {
     trow_token::new(auth_user, tc).map_err(|_| Error::InternalError)
 }
+
+#[options("/login")]
```
amouat, line 132: Do we really need to add options everywhere? Can't we do it in the fairing? I saw https://github.com/SergioBenitez/Rocket/issues/25#issuecomment-313895086

amouat (commented):
I'm not sure about this. Ideally I don't want to add all these OPTIONS methods unless there's no other choice. Can't we handle it in the fairing? I did see some conversation here https://github.com/SergioBenitez/Rocket/issues/25#issuecomment-313895086

amouat:
Also we need an e2e test here. We need to run Trow with CORS on, send an OPTIONS request and see the headers come back. And also verify that they're not on non-options requests or when CORS isn't on.

src/main.rs
```diff
@@ -196,6 +196,11 @@ Must be used with --hub-token or --hub-token-file")
             .help("Location of file with token that can be used for accessing the Docker Hub, used when proxying Docker Hub images")
             .takes_value(true)
         )
+        .arg(
+            Arg::with_name("enable-cors")
+                .long("enable-cors")
+                .help("Enable Cross-Origin Resource Sharing(CORS) requests. Default: false")
```
amouat, line 202: Can you remove the "default false" bit? Also add something like "Used to allow access from web apps (e.g. GUIs).", so that people understand why it's there.

amouat (approved):
Can you update the description for the argument? Just to give users a bit more context about why it's there.

This is excellent though, thanks for the changes - I think it's a lot simpler and better than the last version.

src/main.rs
iamcaleberic: Will do.

Issue #270, opened 2021-08-16
Add CLI arg to set log level


amouat:
At the moment the only way to change the log level is to set the RUST_LOG environment variable. This isn't terrible, but:

 - the prefix "RUST" is confusing; users shouldn't need to know about RUST
 - the variable isn't easily discoverable
 - it's not consistent with everything else which is set via cli args
 
The proposal is to add an argument `--log-level` which sets the log level. RUST_LOG can be left in place for backwards compatability but should be overridden by `--log-level`.

PR #275, opened 2021-09-03
Add CLI parameter for setting the log level

rorymalcolm:
Attempt to address https://github.com/ContainerSolutions/trow/issues/270

While I have _some_ Rust experience this is my first time contributing to a project, your issues section is really well maintained with good 'first issue' labels, so thanks for that :)

Happy to expand on any tests, just didn't seem to be that many great hooks into doing assertions.

- Defaults to "ERROR" to match existing behaviour
- Gives RUST_LOG env var priority to match legacy behaviour
- Documents the valid settings

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -34,6 +34,7 @@ use rocket::fairing;
 use std::env;
 use std::fs;
 use std::path::Path;
+use std::str::FromStr;
 use uuid::Uuid;
 
 mod client_interface;
@@ -93,6 +94,7 @@ pub struct TrowConfig {
     token_secret: String,
     user: Option<UserConfig>,
     cors: bool,
+    log_level: String
 }
 
 #[derive(Clone, Debug)]
@@ -143,7 +145,7 @@ fn init_trow_server(
 }
 
 /// Build the logging agent with formatting.
-fn init_logger() -> Result<(), SetLoggerError> {
+fn init_logger(log_level: String) -> Result<(), SetLoggerError> {
     // If there env variable RUST_LOG is set, then take the configuration from it.
     if env::var("RUST_LOG").is_ok() {
         env_logger::from_env(Env::default().default_filter_or("error")).init();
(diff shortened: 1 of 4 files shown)


amouat:
Cool! Thanks a lot. I'll take a look soon :)

amouat:
Hmm, I'm thinking the CLI arg should have priority over the env var. So:
 - if the arg is explicitly set it should override anything in RUST_LOG
 - if the arg isn't set it should use the RUST_LOG value. I'm hoping this can be done by replacing `unwrap_or("error")` in the code.

 We could also print a warning that both are set and conflicting, but this isn't essential.
 
I understand your point about not wanting to change existing behaviour, but as the default is still to use RUST_LOG, I think we should be ok and it will be less surprising behaviour - I can imagine getting bug reports from users complaining the argument doesn't work as they haven't realised the env var is set.

rorymalcolm:
> I understand your point about not wanting to change existing behaviour, but as the default is still to use RUST_LOG, I think we should be ok and it will be less surprising behaviour - I can imagine getting bug reports from users complaining the argument doesn't work as they haven't realised the env var is set.

Yeah this absolutely makes sense, CLI arg is much more explicitly set by the user, `RUST_LOG` can often be accidentally set elsewhere even though its unlikely to be set in prod it could still make development difficult. What about in the case where neither the CLI flag or `RUST_LOG` are set? Just default to `error` as trow was already doing? Can get this logic fixed in the coming days and will reflag for review.

amouat:
Yes, that sounds perfect!

amouat:
Thanks! I'll take a look soon, but it might be a few days.

rorymalcolm:
> This looks good! I'm wondering if it's worth having a short hand e.g. `-l` for it. I might look at what other software does.
> 
> Apologies for taking so long to get to this.

Thanks, fwiw I really dislike `kubectl`'s `--v` flag syntax, I find it unergonomic and cumbersome, `-ll` is nice imo

Issue #168, opened 2020-09-03
Improve Install Docs


amouat:
The current install docs are a bit all over the place:

 -  The kustomize install instructions are under the `install` directory but the Helm instructions are under `docs`. I guess the best solution is to move the kustomize instructions but leave a placeholder in install pointing to the docs guide.
 - The Helm instructions could do with a bit more context and explanation, especially on TLS ingress and domain names. But some of this is already in the Kustomize instructions, so there might be a way to merge the two.
 - We should consider offering users some context of when to choose between the two. For example, would it be fair to say the Kustomize instructions are better for GitOps and if you need finer control?

pradipta:
On it

awoimbee:
Helm is now the only supported way, it's fully automated (auto generated TLS certs and such)

PR #276, opened 2021-09-21
#168 doc improvements and editing out of date api versions.

KianTigger:
Added missing elements to validate-tmpl.yaml. 

Edited yaml files to get rid of warnings about api versions.

In install/install.md changed numbers after “Assuming your cluster has cert-manager installed:” (not 1-6)

In docker/readme.md first changed command to be ‘docker build -f docker/Dockerfile -t trow ..’ 

Added note in trow/quick-install.md noting that clusterrolebinding must be done every time

### DIFF
--- a/charts/trow/templates/ingress.yaml
+++ b/charts/trow/templates/ingress.yaml
@@ -37,6 +37,7 @@ spec:
             backend:
               service:
                 name: {{ $fullName }}
+                port: {{ $svcPort }}
         {{- end }}
   {{- end }}
 {{- end }}
--- a/docs/KUSTOMIZE_INSTALL.md
+++ b/docs/KUSTOMIZE_INSTALL.md
@@ -32,7 +32,7 @@ resources to run in. Update the YAML under `secretGenerator` with the user name 
 want to use for the registry. 
  3) In the same `kustomization.yaml`, update the domain name to the domain you wish to use for your 
  registry. Update the user name to the user name you set in the previous step. 
- 4) Run `kubectl apply -k install/overlays/mycluster` from the install directory.
+ 4) Run `kubectl apply -k overlays/mycluster` from the install directory.
  5) Set the DNS for your domain to point to the IP for your ingress, which you can find with `kubectl get ingress -n trow-example`. Note that in GKE, 
  this IP is subject to change unless you obtain a static IP.
  It may take a moment for the IP address to populate.
@@ -71,7 +71,7 @@ And also uncomment the following under `patchesJson6902` and modify the domain v
         version: v1
 ```
 
- 2) Run `kubectl apply -k install/overlays/mycluster` from the install directory.
+ 2) Run `kubectl apply -k overlays/mycluster` from the install directory.
 
 It would be better to point Kubernetes at the internal Trow service in step 2, but as this isn't
 running over TLS within the internal network in the default install, we need to use the external
@@ -105,7 +105,7 @@ this.
 
 ## Uninstall
 
-This should be as simple as `kubectl delete -k install/overlays/mycluster`  (changing `install/overlays/mycluster`
+This should be as simple as `kubectl delete -k overlays/mycluster`  (changing `overlays/mycluster`
 to the name of your overlay).
 
 ## Troubleshooting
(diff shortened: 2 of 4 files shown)


docs/KUSTOMIZE_INSTALL.md
```diff
@@ -18,30 +18,30 @@ domain name and any special configuration (e.g. ingress). The overlay will refer
...
 appropriate name.
  2) Open the `kustomization.yaml` file. Change the namespace to whatever namespace you want the Trow
 resources to run in. Update the YAML under `secretGenerator` with the user name and password you 
 want to use for the registry. 
  3) In the same `kustomization.yaml`, update the domain name to the domain you wish to use for your 
  registry. Update the user name to the user name you set in the previous step. 
- 2) Run `kubectl apply -k overlays/mycluster` from the install directory.
- 3) Set the DNS for your domain to point to the IP for your ingress, which you can find with `kubectl
- get ingress -n trow`. Note that in GKE, this IP is subject to change unless you obtain a static IP.
+ 4) Run `kubectl apply -k install/overlays/mycluster` from the install directory.
```
amouat, line 35: this isn't right (if you run it from the install directory, you don't need the install bit)

docs/values.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: networking.gke.io/v1beta1
```
amouat, line 1: what's this file for? Is it meant to be in the docs directory?

quick-install/validate-tmpl.yaml
```diff
@@ -1,9 +1,12 @@
-apiVersion: admissionregistration.k8s.io/v1beta1
+apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingWebhookConfiguration
 metadata:
   name: trow-validator
 webhooks:
   - name: validator.trow.io
+    sideEffects: None
+    admissionReviewVersions:
+      - v1beta1
```
amouat, line 9: what does this mean?

amouat (changes requested):
Some nice small improvements and fixes here.

I'm not sure about adding the "install" prefix everywhere. I think it would be cleaner to say at the start "this assumes you are in the install directory" or similar (same for the "docker" directory thing).

Try to keep lines in the md files to a certain length (it's probably about 100 columns?).

docs/values.yaml
KianTigger: This was the example file you would create if you follow the helm install instructions. Thought it might be useful to just have one created as a template? I can remove it if you think it's not needed.

quick-install/validate-tmpl.yaml
KianTigger: I found the information on them here: https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/

They are required fields so added the simplest and recommended information. Realised I didn't add correct information for admissionReviewVersions (should be ["v1", "v1beta1"])

QUICK-INSTALL.md
```diff
@@ -23,13 +23,11 @@ compatible with containerd based distributions such as microk8s (see #14).
  - You've cloned or downloaded this repo
  - Port 31000 can be reached on the worker nodes. You may need to edit the network policy or firewall settings if running in the cloud. For example, run the following if using GKE:
 ```
-$ gcloud compute firewall-rules create trow \
-    --allow tcp:31000 \ 
-    --description "Allow inbound Trow registry traffic" \ 
-    --project <project name>
+$ gcloud compute firewall-rules create trow --allow tcp:31000 --description "Allow inbound Trow registry traffic" --project <project name>
```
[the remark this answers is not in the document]
KianTigger: Yeah sorry forgot to mention I would keep getting issues like this. (sent a photo on slack)

charts/trow/templates/ingress.yaml
```diff
@@ -35,8 +35,8 @@ spec:
         {{- range .paths }}
           - path: {{ . }}
             backend:
-              serviceName: {{ $fullName }}
-              servicePort: {{ $svcPort }}
+              service:
+                name: {{ $fullName }}
```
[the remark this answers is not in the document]
KianTigger: I can't remember now but I do remember making a decision to remove the port. I think it might have not worked if I left the port in as $svsPort wasn't valid? But I can't remember how I tested that file anymore unfortunately sorry. 
I've added port: {{$svsPort}} back in and quick install and helm and kustomize seem to work fine.

PR #284, opened 2021-12-27
Update link to kustomize install.

ariyonaty:


amouat:
Good catch, thanks very much!

Apologies for the slowness in merging - I was away over the holiday period.

PR #302, opened 2022-01-20
Correct off by one error with the range header.

bglimepoint:
As per https://github.com/moby/buildkit/issues/2568 and https://github.com/docker/for-linux/issues/1296, the header being off by one is an issue when using BuildKit's buildctl command as the manifest contains the wrong size - being one too large which then fails when trying to read one byte too many from the image.

The issue can be seen when viewing the manifest in the registry data:

In the layers payload:

```
  ...
  "size": 894478744,
  "digest": "sha256:929d4d417adbdafe947076e47161e10f88afcb63716ff7ec3205bc8f62be085f"
```

Whilst a `du -b` of the blob `929d4d417adbdafe947076e47161e10f88afcb63716ff7ec3205bc8f62be085f` is actually `894478743`.

The error doesn't present when using the Docker daemon to build (using BuildKit or not) as it doesn't check this size. It needs to be the buildctl directly.

Note that buildctl caches the manifest data from the first place the image was pulled from, so pulling from a broken registry and then an unbroken one; the build will still fail. The error in buildctl is:

```
/moby.buildkit.v1.Control/Solve returned error: rpc error: code = Unknown desc = failed to compute cache key: failed to read expected number of bytes: unexpected EOF
```

Or

```
error: failed to solve: failed to compute cache key: failed commit on ref "layer-sha256:929d4d417adbdafe947076e47161e10f88afcb63716ff7ec3205bc8f62be085f": "layer-sha256:929d4d417adbdafe947076e47161e10f88afcb63716ff7ec3205bc8f62be085f" failed size validation: 894478743 != 894478744: failed precondition
```

Depending on the version of buildctl.

---

Having never written any Rust, this is probably not the best way to solve this :-D So feel free to adjust as needed - but I wanted some code to test with. With this change buildctl now builds correctly.

I think the change to `upload_info.rs` fixes the issue we saw, but saw the repetition of the issue so changed it too.

Also, thanks for all the work you do with Trow, it's been great for us :-)


amouat:
Ah, I've hit this before! I wondered what was going on - I ended up having to edit manifests directly.

bglimepoint:
(Just repushed with a formatting fix that I hadn't notice (which was why the tests failed))

amouat:
Let me see if I can come up with a test for this and I'll get it merged as soon as I can. I'm a little busy today, so it might be Monday.

amouat:
Looking at this now. You're absolutely right about the problem, but I think it's probably better to fix on the `ContentInfo` level, so that the responses don't need to change the data. I should even be able to put in a check that the max range < length.

amouat:
Fixed in #304

Big thanks for the report, apologies for not using the PR.

amouat:
Also, please test and open an issue if you hit any problems.

bglimepoint:
> Fixed in #304
> 
> Big thanks for the report, apologies for not using the PR.

Absolutely no issue at all :-) I can definitely see how fixing it in blob is the better solution :+1:

PR #304, opened 2022-01-27
Fix ranges

amouat:
As per #302, range handling was broken in some parts of Trow. (Strangely enough, it was correct in some other parts!)

This fixes the problem at a slightly different level to #302 (where the ranges are created) and adds tests.


amouat:
@bglimepoint hopefully this will fix things and the tests will stop it breaking again in future.

You can test by checking out the source, but there's currently a small problem with Docker builds that I'll try to fix next.

bglimepoint:
Thanks @amouat!

Absolutely no worries, things take as long as they take :-)

PR #307, opened 2022-01-28
quick-install: Compatibility fixes for Kubernetes v1.22+

### LINKED ISSUE #303: CSR was moved to V1 api in K8s version > 1.18
**Describe the bug**
Starting in K8s version 1.19 the CSR resource is not in the beta api version anymore (https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/). This will lead to a crash of the init-container of the trow-deployment if using the quick start installation. 

**To Reproduce**
Use a cluster with a recent kubernetes version and run the quick installation. The main script will wait forever for the CSR to be signed. The init-container will crash because kubernetes will reject the CSR resource caused by the wrong api version. 
In addition the field signerName is missing. 

**Expected behavior**
Init-container should not crash and create a valid CSR resource. I guess in addition a k8s version check should be implemented.

**Output/Logs**
```
k logs -n trow trow-deploy-f6cb88bc7-c69sd -c trow-init
Mon Jan 24 12:23:01 UTC 2022
Error from server (NotFound): secrets "trow-tls" not found
Generating new certificate
Getting IP of trow service
POD NAMESPACE: trow
POD NAME: trow-deploy-f6cb88bc7-c69sd
POD IP: 10.244.3.149
SERVICE IP: 10.111.125.45
2022/01/24 12:23:01 [INFO] generate received request
2022/01/24 12:23:01 [INFO] received CSR
2022/01/24 12:23:01 [INFO] generating key: rsa-4096
2022/01/24 12:23:03 [INFO] encoded CSR
Error from server (NotFound): certificatesigningrequests.certificates.k8s.io "trow.trow" not found
error: unable to recognize "STDIN": no matches for kind "CertificateSigningRequest" in version "certificates.k8s.io/v1beta1"
```

**Trow Info**
 - Install method: quick-install
 - Version/Image Name: commit 9925d57dae8095a80c99e997e9470ed0be13c425

**Kubernetes**
 - Kubernetes distro/host: Ubuntu 20.04.03; kubeadm
 - Kubernetes version: v1.23.1
 - Container Runtime: docker

**Additional context**
A valid resource could look like this: 
```bash
  cat <<EOF | kubectl create -f -
apiVersion: certificates.k8s.io/v1
kind: CertificateSigningRequest
metadata:
  name: trow.$POD_NAMESPACE
spec:
  groups:
  - system:authenticated
  request: $REQ
  signerName: kubernetes.io/kube-apiserver-client
  usages:
  - digital signature
  - key encipherment
  - server auth
EOF
```

MukeshS-hexaware:
Fixes #303.

### DIFF
--- a/quick-install/trow-init-pod/csr.yaml
+++ b/quick-install/trow-init-pod/csr.yaml
@@ -1,4 +1,4 @@
-apiVersion: certificates.k8s.io/v1beta1
+apiVersion: certificates.k8s.io/v1
 kind: CertificateSigningRequest
 metadata:
   name: trow.kube-public
@@ -6,8 +6,8 @@ spec:
   groups:
   - system:authenticated
   request: $(cat server.csr | base64 | tr -d '\n')
+  signerName: kubernetes.io/kubelet-serving
   usages:
   - digital signature
   - key encipherment
   - server auth
-EOF
--- a/quick-install/trow-init-pod/init.sh
+++ b/quick-install/trow-init-pod/init.sh
@@ -46,11 +46,16 @@ else
     "$POD_IP",
     "$SERVICE_IP"
   ],
-  "CN": "trow.$POD_NAMESPACE",
+  "CN": "system:node:trow.$POD_NAMESPACE",
   "key": {
     "algo": "rsa",
     "size": 4096
-  }
+  },
+  "names": [
+        {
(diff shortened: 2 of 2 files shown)


amouat:
This looks good! Thank you.

amouat (approved):

Issue #309, opened 2022-02-01
Race conditions on the file system


awoimbee:
**Describe the bug**
Concurrently pulling images that share layers will lead to corrupted data. The major culprit is https://github.com/ContainerSolutions/trow/blob/b7de218b810e7c03bbcb81a1968af00c236de2d3/lib/server/src/server.rs#L424

There are many calls to `File::create(Path)` without any check, these calls truncates existing files and will corrupt them if multiple tasks are writing concurrently.

**To Reproduce**
No reproduction, this is more an epic than a standard issue.

**Expected behavior**
No filesystem corruption due to race conditions.

amouat:
Crap. Thanks for finding this.

amouat:
I just had a quick look - note the use of UUIDs, so I'm not sure this is a problem in practice?

PR #314, opened 2022-02-01
Add multiplatform (manifest list) proxying + fix concurrent writes to files + add concurrent download of layers

### LINKED ISSUE #310: Trow cannot proxy multiplatform images
**Is your feature request related to a problem? Please describe.**
https://github.com/Extrality/trow/issues/1
This missing feature causes bugs when proxying registries other than dockerhub, and dockerhub might drop compatibility with older clients at any moment.
Example error message:
```
[2022-01-19T10:44:34Z WARN  trow_server::server] Internal error with manifest ErrorMessage { msg: "Digest sha256:1cbb4dc8437d49f51a11bdbbebff3bd494a1f81a40deedd93db2c9af9d433be1 not in repository f/ghcr/extrality/cert-manager-webhook-namecheap" }
```

It also seems to cause weird issues with *some* images from dockerhub but I don't have a good test case.
Trow currently has arm builds/releases but can't proxy arm docker images (dockerhub falls back to amd64 since Trow doesn't have the Accept header for `manifest.list` manifests).

**Describe the solution you'd like**
Trow can proxy multi-platform container images.

awoimbee:
Sorry, I wanted this PR to be smaller but then went down the rabbithole of file integrity and concurrent writes.

- Fixes #310 (Fixes https://github.com/Extrality/trow/issues/1): 
    Relevant changes are in `fn create_accept_header` and `async fn download_manifest_and_layers` + newly created `async fn download_blob`
    We have to download every layer of every platform because a client might ask for any random layer.
- Lots of improvements regarding #309:
    I ended up creating `lib/server/temporary_file.rs`, take a look at `test_temporary_file_async_cancellation` ;)
    Thanks to this wrapper:
    - there are no stale scratch file: either a file exists and is being worked on, or it doesn't exist
    - trow cannot write to a scratch file without being its creator: no concurrent writes
- Concurrent download of multi platform manifests and of layers: the speed increase is noticeable
- I didn't understand why Trow used `fs::copy` to move blobs from the scratch to their destination, so I changed it to `fs::rename` (I think it's fair to assume that both these directories are on the same filesystem), the speed increase is noticeable here too

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -92,6 +92,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-recursion"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2cda8f4bcc10624c4e85bc66b3f452cca98cfa5ca002dc83a16aad2367641bea"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
 name = "async-stream"
 version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2650,6 +2661,7 @@ dependencies = [
 name = "trow-server"
 version = "0.3.5"
 dependencies = [
+ "async-recursion",
  "async-stream",
  "bytes",
  "chrono",
--- a/lib/server/Cargo.lock
+++ b/lib/server/Cargo.lock
@@ -33,6 +33,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "84450d0b4a8bd1ba4144ce8ce718fbc5d071358b1e5384bace6536b3d1f2d5b3"
 
 [[package]]
+name = "async-recursion"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2cda8f4bcc10624c4e85bc66b3f452cca98cfa5ca002dc83a16aad2367641bea"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
 name = "async-stream"
 version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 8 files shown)


splitice:
wow, great work with the filesystem corruption points. Thats a serious bug.

I'm not a maintainer but I'm not sure I agree with "I think it's fair to assume that both these directories are on the same filesystem". It might be the case 90% of the time, but thats not a good argument. Perhaps a default and a flag? Or attempt a copy if the rename fails?

awoimbee:
@splitice keep in mind that the cli [accepts a `--data-dir` arg](https://github.com/ContainerSolutions/trow/blob/aff7336c93250903044d10f1128a37326f50c0c7/src/main.rs#L60), this dir then contains `blobs/` `manifests/` `scratch/`. So the file operations are all happening inside the `data-dir` (not between eg `/tmp` and `/data`).

splitice:
While I see your point @awoimbee I much prefer retaining compatibility and flexibility. Is there some reason attempting a move if the rename fails would be difficult or suboptimal?

The errno could be checked.

awoimbee:
Anything can be done, the question is should it be done. I forgot to add that `fs::rename` is [an atomic operation](https://rcrowley.org/2010/01/06/things-unix-can-do-atomically.html) while `fs::copy` is not, meaning `fs::copy` might cause Trow to serve invalid files.
I also don't want to spend more time on this, nor do I want to write tests for edge cases (eg `test_copy_if_rename_fails`). So if @amouat says I need to roll-back, I'll roll-back, otherwise not.

amouat:
Just looking at this now. Thanks a lot for all the work here.

amouat (approved):

amouat:
Yeah, I think rename is better. I am worried there was a reason I didn't do it, but I suspect I just never noticed. This way makes things a lot simpler. Thanks for the tests as well, this was a great PR.

@splitice thanks for bringing up those points; they both need to be thought about. In this case, I really don't see much impact on existing users. And @awoimbee is correct about the filesystem. This does make it harder to move to a system where those directories are on a different FS, but I don't expect that to be a requested feature in the short term at least. And if we're wrong we can always rework this.

PR #319, opened 2022-03-14
Allow overriding the image tag.

bglimepoint:
To allow consumers to modify it if needed.

No worries if you don't want to allow this :-) But we noticed that the charts hadn't been updated as part of the 0.3.5 release and hence noticed that this couldn't be overridden, and it seemed useful as an option :-) Especially as the image itself can be overridden :-)

### DIFF
--- a/charts/trow/templates/statefulset.yaml
+++ b/charts/trow/templates/statefulset.yaml
@@ -22,7 +22,7 @@ spec:
     {{- end }}
       containers:
       - name: trow-pod
-        image: "{{ .Values.image.repository }}:{{ .Chart.AppVersion }}"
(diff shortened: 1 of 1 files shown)


bglimepoint:
Oh, it is there and I'm blind :-D But still, didn't seem un-useful? :-)

amouat:
Looks good to me and makes sense as far as I'm concerned. Personally, I use Kustomize more than Helm, so I struggle with it sometimes. I'm a little unsure why there's two chart files to be honest. For the moment I'll update them both to be the same.

Can you just confirm that index.yaml on GitHub pages _does_ have the correct image?

amouat (approved):

PR #321, opened 2022-04-16
update deps: replace failure by anyhow

### LINKED ISSUE #160: Replace failure crate
The codebase makes extensive use of the failure crate, but this is now the old way of doing things following changes to the way Errors work in Rust. We may be able to use another crate e.g. [anyhow](https://lib.rs/crates/anyhow) or just use standard features.

Looking at the crate documentation, it looks like anyhow is the correct way to go:

"The anyhow::Error type works something like failure::Error, but unlike failure ours is built around the standard library's std::error::Error trait rather than a separate trait failure::Fail. The standard library has adopted the necessary improvements for this to be possible as part of RFC 2504."

awoimbee:
Fixes #160

Hi, I replaced `failure` by `anyhow` and while I was at it, updated the other deps.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -152,49 +152,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
-name = "axum"
-version = "0.5.1"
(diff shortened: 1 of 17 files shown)


amouat:
Oh wow, this looks good. Thanks! I'm a bit limited in time atm but I'll try to take a look soon.

amouat (approved):

PR #323, opened 2022-04-16
remove calls to File::create

### LINKED ISSUE #309: Race conditions on the file system

awoimbee:
Fixes #309

To be fair, this issue was fixed by  #314.
This PR only removes the remaining calls to the (very) evil `File::create` method.

-> regarding `save_tag`, `append` type writes are atomic on most filesystems ;)

### DIFF
--- a/lib/server/src/server.rs
+++ b/lib/server/src/server.rs
@@ -4,7 +4,7 @@ use std::collections::HashSet;
 use std::fmt;
 use std::fs::{self, DirEntry, File};
 use std::io;
-use std::io::{BufRead, BufReader, Read, Write};
+use std::io::{BufRead, BufReader};
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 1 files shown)


amouat (approved):

amouat:
Needs a rebase, but this looks good.

PR #324, opened 2022-04-19
use cargo workspace

awoimbee:
Cargo workspaces have been the standard way to architecture a multi-crate project for a few years now: https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html

I think this project was created before cargo workspaces were a thing. It's time to upgrade !

Amongst the improvements:
- fixes rust-analyzer errors (caused by `import!()` a file outside the rust project)
- seems to make compilation faster.
- when running `cd trow-server && cargo build`, the central `Cargo.lock` will be used (no more stray `lib/server/Cargo.lock`)


amouat:
I'll try to take a look at this midweek.

amouat:
This does seem a lot cleaner!

amouat (approved):

amouat:
Let's get the other two merged then we can rebase this. Looks good though!

PR #328, opened 2022-06-10
fix memory usage: use streams to write blobs

### LINKED ISSUE #327: Large memory usage spike when pulling big layers
I'll make it short:
When pulling big images like `trow.example.com/f/docker/nvidia/cuda:11.0.3-devel-ubuntu18.04`, trow downloads and writes image layers using `file.write_all(&resp.bytes().await?).await?;`. This is an issue because if a layer is 1GB, Trow will use 1GB of RAM before dumping everything to disk.
Offending code: https://github.com/ContainerSolutions/trow/blob/22fc316874b087d7a25a5b549db7f84b52e378a2/trow-server/src/server.rs#L469

awoimbee:
Fixes #327.

Note:
I'm currently working on https://github.com/ContainerSolutions/trow/tree/multi-proxy-registry, a spiritual successor to https://github.com/ContainerSolutions/trow/pull/250 with the goal to also integrate something like https://github.com/phenixblue/imageswap-webhook so users don't have to edit every image field of every deployment to use Trow.
Are you open to such change @amouat ?

### DIFF
--- a/trow-server/Cargo.toml
+++ b/trow-server/Cargo.toml
@@ -26,7 +26,7 @@ thiserror = "1.0"
 serde_derive = "^1.0"
 trow-protobuf = { path = "../trow-protobuf" }
 rustc-serialize = "0.3"
-reqwest = { version = "0.11", features = ["json"] }
+reqwest = { version = "0.11", features = ["json", "stream"] }
 prometheus = { version = "0.13"}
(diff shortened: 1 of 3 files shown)


amouat (approved):
Looks good!

I haven't tested, but this sort of thing should be good as long as the tests pass.

PR #332, opened 2022-10-27
Add ability to override securityContext + automountServiceAccountToken

sarthur-limepoint:
We've started using some K8s security scanners and they have suggest that the security posture of pods should be hardened from the defaults.

This allows the pod and container `securityContext` objects as well as the pod `automountServiceAccountToken` field to be modified when using the Helm chart.

Since Trow does not appear to use the Kubernetes API at present, this also sets the `automountServiceAccountToken` to false by default.


amouat (approved):

amouat:
Thanks! LGTM. Apologies for wait on merge.

Issue #333, opened 2022-12-10
Trow unreachable over IPv6


YannikSc:
**Describe the bug**

When running a cluster with DualStack/IPv6 trow is not reachable as it only listens on `0.0.0.0` (all IPv4 interfaces).

This could be mitigated by letting Trow/Rocket listen by default on `::` (actually all interfaces IPv4 and IPv6). 

Another workaround would be to allow the helm chart to set the `--host` flag.

**To Reproduce**

- Setup a cluster with IPv6 support
- Setup an ingress controller with IPv6 support
- Setup trow + ingress (or IPv6 service)

**Expected behavior**

Trow is reachable at the configured domain through the ingress/directly via IPv6

**Trow Info**
 - Install method: Helm
 - Helm Chart info: 
```
apiVersion: v2
appVersion: 0.3.5
description: Helm chart for Trow registry
name: trow
type: application
version: 0.3.5 
```

**Kubernetes**
 - Kubernetes distro/host: kubeadm on Arch
 - Kubernetes version: v1.25.5
 - Container Runtime: ContainerD

awoimbee:
Note that Trow uses Rocket, which does not support listening to multiple addresses in a single instance. The workaround described here might work tho: https://github.com/SergioBenitez/Rocket/issues/1581.
Otherwise the solution is to have multiple instances of rocket running (like I did [here](https://github.com/ContainerSolutions/trow/pull/329)).

amouat:
Thanks @awoimbee !

PR #334, opened 2022-12-10
Allow `--host` to be set through helm

YannikSc:
This enable optional IPv6 support by allowing trow to be bound to `::` through the helm chart

Related to #333. 

I don't wanted to create a "core change" for that by changing the default value for the CLI arg, as this might also break windows support (depends on how windows handles `::` when binding sockets), or possibly even non-standard Linux configurations.

I tested it successfully in my test cluster and it works.


amouat (approved):

amouat:
Can you rebase? I commented out this test in the latest code.

amouat:
Great, thanks for the PR!

PR #346, opened 2023-07-05
Update a bunch of stuff.

amouat:


amouat:
@awoimbee I wanted to merge this before moving, but I no longer have the ability to force merges without review :)

awoimbee:
I think this PR should only update your email, not a fan of adding unused imports and `//TODO: tests are interleaved to avoid some race condition that should be investigated` while this race condition does not seem to happen in  https://github.com/Extrality/trow

amouat:
I was mainly trying to update a bunch of dependencies to resolve vulnerabilities. But then the tests didn't pass, so I needed to make some changes.

The alternative is to first merge the extrality fork, would you rather do it that way around?

awoimbee:
Sorry, missed this one

> The alternative is to first merge the extrality fork, would you rather do it that way around?

Absolutely, my fork has a lot of changes, rebasing would not be fun !

Issue #339, opened 2023-01-28
Proxy registry.k8s.io and k8s.gcr.io, and ghcr.io the same as docker.io


mysticaltech:
**Is your feature request related to a problem? Please describe.**

Hello, we would like to you this for our project [kube-hetzner](https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/) where some nodes in the clusters that are deployed with our tools on Hetzner Cloud are blacklisted by the security solution (Cloud Amor) that gcr based registries, including registry.k8s.io and k8s.gcr.io, and ghcr.io use. These nodes get 403s when trying to pull images leading to subsequent ImagePullBackoff and ErrImagePull errors.

**Describe the solution you'd like**

You solution is beautiful, it has everything we need, almost. As we can expose the ingress through the LB that is usually not blacklisted, or even request a new one. 

But for that to work, we could need you to allow proxying these registries too, specifically `registry.k8s.io` and `k8s.gcr.io`, and `ghcr.io`, not only docker.io.

**Describe alternatives you've considered**

We've considered jfrog and gitlab, but also harbor. But none are as good as what you propose here! As it does not demand an extra server.

**Additional context**

You are the solution to a long-standing problem, if this work, we are saved and trow will be used by hundreds of clusters, if not tens of thousands over time, as we plan to add you as an addon to kube-hetzner.

Issue #330, opened 2022-06-22
manifests for /f/docker/library/xx and /f/docker/xx are not saved at the same place


awoimbee:
https://github.com/ContainerSolutions/trow/blob/0b20a56ad8fe1195cbde2d7aa7c9910f448f7ae9/trow-server/src/server.rs#L524

`local_repo_name` is not validated/parsed, meaning manifests for `/f/docker/library/alpine` and `/f/docker/alpine` won't be located at the same place (so won't be equal).

PR #329, opened 2022-06-17
Proxy any registry and do it automatically (using MutatingWebhookConfiguration)

awoimbee:
Sorry for the size of this PR, I saw no other way to do this (and do it fast).

The goals of this PR:
- Fix the helm chart
- One click install of Trow with its admission webhooks through helm
- Add a mutation admission webhook so images in pods are automatically routed through Trow => Fixes https://github.com/ContainerSolutions/trow/issues/322
-  Ability to proxy any registry (#250) -> handle multiple types of authentication methods (https://github.com/ContainerSolutions/trow/pull/279)

Other changes that happened:
- replace the validation configuration: the old way of configuring validation was way too complicated for me. There were 6 CLI args + the logic regarding if the image is "local" or not. => replaced by a `.yaml` that only configures the webhook (no side effects) 
- move some logic from trow-server to modules: `proxy_auth.rs`, `image.rs` => it made writing tests so much easier
- use `kube` and `k8s-openapi` instead of custom types for `AdmissionReview`

Deprecation of `./quick-install/` and `./install/`:
I want to greatly simplify the deployment by only providing an easy to use helm chart. These folders push the user to run many manual steps/scripts and they duplicate deployment config. I did not outright delete these folders because this PR has already too many changes, but it's planned.

Notes:
- I am running this Trow version in prod. Deployed via Helm, chart hosted at: https://github.com/awoimbee/helm-charts/releases/tag/trow-0.4.1
- The [helm_chart_validation](https://github.com/ContainerSolutions/trow/runs/6932972845?check_suite_focus=true#logs) test fails because it uses the old docker container, which takes different CLI args from the new one. This PR is already big enough as is so I'm not gonna fix that.
- the webhook patch deployment thing is largely inspired by ingress-nginx

PR #279, opened 2021-11-25
Proxying support from extrality fork

amouat:
Looking at how to merge changes from @awoimbee https://github.com/Extrality/trow/

At the moment this PR is just so I can see the changes and start refactoring into something that could be merged.

Stuff will be pulled out and added in later PRs.

PR #347, opened 2023-09-18
merge Extrality/trow into trow-registry/trow

awoimbee:
This includes all the changes/improvements made in https://github.com/Extrality/trow (see https://github.com/Extrality/trow/pulls).

Closes #345, #343, #340, #329, #279, #250, #339, #330, #229, #189, #168


amouat (approved):

amouat:
I'd really like to re-enable cosign. I don't really see the problem with signatures.

awoimbee:
Thanks !

>  cosign and no-tls were pretty useful I thought

* `cosign` may be added back later but it needs some work, I want to avoid polluting the package page with signatures like what happened in https://github.com/Extrality/trow/pkgs/container/trow.
* `no-tls` is still there: before, trow looked for certificates in `/etc` unless `--no-tls` was given. Now you need to use the `--tls` option, which allows passing arbitrary paths to certificates

amouat:
Ok, great. Very happy to hear about tls :)

I should manage to look into cosign at some point. I might also manage to look into simplifying the docker build. Both of these things have cross-over with my current work.

docs/KUSTOMIZE_INSTALL.md
jglick, line 1: Also broke https://github.com/Trow-Registry/trow/blob/main/docs/USER_GUIDE.md → https://github.com/Trow-Registry/trow/blob/main/docs/KUSTOMIZE_INSTALL.md link

PR #353, opened 2023-09-19
docs: update helm registry URL

adrienthebo:


awoimbee:
I think https://trow.io points to https://trow-registry.github.io/trow.
I have no vision on the domain name (for how long it's valid, how much it costs, who owns it, ...), so I'll let @amouat handle this one.

adrienthebo:
My shell history reveals that my `helm repo add` invocation was incorrect, this can be closed.

PR #364, opened 2023-10-28
Remove gRPC (part 1)

awoimbee:
This PR removes gRPC from the Trow code, thus reducing the maintenance burden.
Now more changes will be needed to fully reconciliate trow and trow-server, and to make the code nice.

It's been a while since I have not updated `tonic` nor `prost`, removing them was easier, reduces memory usage, removes the double interface pain (`--HTTP-->trow--gRPC-->server` becomes `--HTTP-->trow`), improves logging, and will improve passing errors from trow-server to trow.

The original plan for trow (HA, multi server) is still possible that way too !


awoimbee:
I saw you commented my other PR, WDYT about this one @amouat ?

PR #365, opened 2023-10-28
Replace manual wilcard path segment handling by new macro: endpoint_fn_7_levels

awoimbee:
Many routes of the CRI look like `/v2/:image_name*/blobs/:digest`.
Axum doesn't support wildcard segments in the middle of a route (rocket didn't either).
The PR replaces manually writing every *_2level, *_3level, etc function by a macro invocation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1986,6 +1986,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "paste"
+version = "1.0.14"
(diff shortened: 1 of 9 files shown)


amouat (approved):
Cool.

It's still annoying but much better. The spec isn't fantastic here tbh.

awoimbee:
It will be much easier to maintain without having to duplicate every change 7 times !

(The tests all pass locally, I will fix little incompatibilities with the gh actions runners later)

PR #382, opened 2024-05-13
Refacto: separate storage backend, merge client_interface and trow_server into "registry"

awoimbee:
I've been working on this for a while...

Initial goal: separate all storage operations from trow_server so features such as tiered storage and s3 backend could be added later.

Final state:
* storage operations are united in `storage.rs`
* tests don't require neither subprocesses nor to be ran sequentially anymore
* `trow_server` and `registry_interface` are united as `registry`
* update deps

I think the next step is to add a database, working around the "manifest_history" files is not straightforward.
This PR breaks compatibility with the `data` dirs of old trow versions.


awoimbee:
I want to get this over with, I will fix the OCI compliance tests later

amouat:
Yeah go for it. I don't really have time to review atm unfortunately.

PR #384, opened 2024-10-11
webhook tls certs: option to create via certmanager or existingSecretRef

awoimbee:
This allows deploying Trow without having to create a ClusterRole for the webhook patch logic.

To keep things tidy, all resources defined in `templates/webhooks/patch/` are now inside `templates/webhooks/tls-patch.yaml`.


charts/trow/templates/webhooks/tls-certmanager.yaml
```diff
@@ -0,0 +1,62 @@
...
+  namespace: {{ .Release.Namespace }}
+spec:
+  ca:
+    secretName: {{ include "trow.fullname" . }}-cm-admission-root-cert
+---
+# generate a server certificate for the apiservices to use
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: {{ include "trow.fullname" . }}-admission
```
jerome-blanche, line 44: @awoimbee Do you want to use `-webhooks` as the suffix, to match what is used in the annotations in the *webhook.yaml files?

PR #392, opened 2025-01-12
Point button to Helm Install

albal:
Used to point to install in root of repo - it no longer exists so probably best to point to the Helm Installation page.


awoimbee (approved):
Thanks

PR #394, opened 2025-01-15
Store metadata in sqlite

### LINKED ISSUE #145: Create Internal Data Structure / Cache
At the moment the file system is effectively the database and source of all truth for Trow.

I think it's important that all data is captured in the filesystem, allowing Trow to be easily copied, restarted, backed-up etc. But there needs to be an internal in-memory data-structure to allow fast updates, statistics, metrics, checks etc. 

This was something that was discussed a lot in the early days of trow, when the overall design was being discussed. The hope was that we would be able to use CRDTs to support a distributed Trow network with multiple Trow backend instances. Each instance may have a different set of images and would need to coordinate with peers, hence the CRDT. It's important to keep such use cases in mind as we build out the data structure, so that it remains possible to implement them. The CRDT may or may not be part of the internal data structure but I think it's relevant to think about here.

Another thought I had was that the contents of the registry can be expressed as a SHA of the manifests directory. Each subdir should have it's own SHA and the manifest directory then has a sha of its subdirs. (I later realised I was describing a sort of merkle tree). Keeping these checksums allows us to verify that state hasn't changed unexpectedely. For example it could be written out on registry shutdown and used to verify contents when restarted (or there could be random/requested verification runs).

The blobs directory should also have a SHA representing all the contents. We can then do the following:

    - get all blob SHAs from _manifests_ and munge into single SHA ("all manifest digest") somehow (would require ordering)
    - get all SHAs from _blob folder_ and munge into single SHA ("all blob digest")
    - if they match, we know the registry is in sync (no dangling blobs).
    - we should also verify each SHA matches the blob

The SHAs should be returned from an endpoint as the current repo state. Verify command could be triggered by API or on startup.

### LINKED ISSUE #391: [Bug]: Can't login to Trow registry when secured with password
### Describe the bug

I deploy trow on a GKE cluster using Helm chart. When I set user/password to protect it, I can't `docker login` to list/push/pull images:
```
kuba@***:~$ docker login -u <correct user> -p <correct password> trow.*****
WARNING! Using --password via the CLI is insecure. Use --password-stdin.
Error response from daemon: login attempt to https://trow.*****/v2/ failed with status: 401 Unauthorized
```

However, when I give wrong credentials an error is different:
```
kuba@***:~$ docker login -u <correct user> -p <incorrect password> trow.*****
WARNING! Using --password via the CLI is insecure. Use --password-stdin.
Error response from daemon: Get "https://trow.image-proxy-test.respo.tech/v2/": unauthorized: authentication required
```

I'm sure that the network setup (Service, Ingress, TLS cert) is fine - when I comment out user/password in `values.yaml` all works fine - I can list/push/pull images.

I also checked the K8s password Secret value and it's mount as a file in the Trow container - they are fine. The file's content is the same as the password I enter into `values.yaml`.  

### To Reproduce

GKE cluster version: 1.30.x
GKE node pools version: 1.30.x
trow chart version: 0.8.1
trow image tag: 0.6.4

trow is set up with Google Managed Certificates on GKE as in HELM_INSTALL.md

`values.yaml` content (password and domain masked):
```
# Default values for trow.
# This is a YAML-formatted file.
# Declare variables to be passed into your templates.

replicaCount: 1

image:
  repository: ghcr.io/trow-registry/trow
  tag:
  pullPolicy: Always

trow:
  ## if using NodePort, this can be set to 127.0.0.1:XXXX
  ## Where "XXXX" is the value of service.nodePort
  domain: trow.*****
  user: user
  password: *****

  ## The Trow validation webhook validates which images are allowed to run on your cluster.
  ## Note: it runs AFTER the mutating webhook.
  validationWebhook:
    enabled: true
    failurePolicy: Ignore
    ## The validation webhook need to be configured.
    ## Note: it runs AFTER mutation webhooks.
    ## Example:
    config:
      ## Default action for unknown images (Deny or Allow)
      default: Allow
      ## List of allowed image URI prefixes
      allow:
        - 0.0.0.0/0
      deny: []
        # - 127.0.0.1/f/docker/nope/
  ## The proxyRegistries section allows you to configure Trow to act as a proxy for other registries.
  proxyRegistries:
    ## The Trow mutating webhook will automatically rewrite the pod image URLs to use the Trow proxy.
    webhook:
      enabled: true
      failurePolicy: Ignore
    config:
      # offline: false
      registries:
        - alias: docker
          host: registry-1.docker.io
      #   - alias: ghcr
      #     host: ghcr.io
      #     ignore_repos: ["trow-registry/trow"]
  ## For more info on log levels see https://docs.rs/tracing-subscriber/0.3.17/tracing_subscriber/filter/struct.EnvFilter.html
  logLevel: info

## Job to configure the webhooks
webhooks:
  ## It is possible to restrict webhook access to some namespaces, ref:
  ##   https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-namespaceselector
  # namespaceSelector:
  replicas: 2
  ## Config to create a TLS cert for the webhooks (required if webhooks are used)
  tls:
    ## Create the cert using certmanager
    certmanager:
      enabled: false
      rootCert:
        # default to be 5y
        duration: ""
      admissionCert:
        # default to be 1y
        duration: ""
      # issuerRef:
      #   name: "issuer"
      #   kind: "ClusterIssuer"
    ## Create a self-signed cert using kube-webhook-certgen to patch the webhooks (creates ClusterRole)
    patch:
      enabled: true
      image:
        image: registry.k8s.io/ingress-nginx/kube-webhook-certgen:v1.4.4@sha256:a9f03b34a3cbfbb26d103a14046ab2c5130a80c3d69d526ff8063d2b37b9fd3f
        pullPolicy: IfNotPresent
    ## Use an existing secret that contains tls.crt and tls.key
    existingSecretRef: ~

imagePullSecrets: []
nameOverride: ""
fullnameOverride: ""

service:
  type: NodePort
  port: 8000
  ## Specify the nodePort value for the LoadBalancer and NodePort service types.
  nodePort:

ingress:
  enabled: true
  gke: true
  annotations:
        networking.gke.io/managed-certificates: trow-certificate
    # kubernetes.io/ingress.class: "gce"
    # kubernetes.io/ingress.class: nginx
    # kubernetes.io/tls-acme: "true"
  hosts:
    - paths: ["/"]
      # use "none" to not set a host (otherwise defaults to trow.domain)
      host:
  tls: []
  #  - secretName: chart-example-tls
  #    hosts:
  #      - chart-example.local

resources:
  limits:
    memory: 256Mi
  #   cpu: 100m
  # requests:
  #   cpu: 100m
  #   memory: 128Mi

# run pods on dedicated, persistent nodes (other nodes are in autoscaling groups with min number of instances set to 0)
nodeSelector:
  *****.cluster.node.type: infra-services

tolerations:
  - key: components.gke.io/gke-managed-components
    operator: Exists
    effect: NoSchedule

affinity: {}

volumeClaim:
  accessModes: ["ReadWriteOnce"]
  resources:
    requests:
      storage: 40Gi

```

### Expected behavior

I should be able to protect Trow registry with user and password and be able to log in with proper credentials.

### Output/Logs

During failed `docker login` attempts (the ones with proper credentials) I can see in trow pod logs:
```
2025-01-09T22:33:46.906117Z  INFO request{method="GET" path="/v2/"}: trow::routes: done in 54.535µs
2025-01-09T22:33:47.012424Z  INFO request{method="GET" path="/login"}: trow::routes: done in 18.573506ms
2025-01-09T22:33:47.137624Z  WARN request{method="GET" path="/v2/"}: trow::response::trow_token: Failed to decode secret: Base64 error: Invalid byte 45, offset 8.
2025-01-09T22:33:47.137675Z  INFO request{method="GET" path="/v2/"}: trow::routes: done in 159.229µs
```

With improper credentials logs are different:
```
2025-01-09T22:37:10.743599Z  INFO request{method="GET" path="/v2/"}: trow::routes: done in 53.722µs
2025-01-09T22:37:10.846017Z  INFO request{method="GET" path="/login"}: trow::routes: done in 18.306034ms
```

### Trow Info

trow chart version: trow-0.8.1
trow image tag: 0.6.4

### Kubernetes

- GKE cluster version: 1.30.x
- GKE node pools version: 1.30.x
- Container Runtime: containerd://1.7.24

### Additional context

Trow seems great and it covers my use case perfectly but I can't use it when exposed unprotected to the Internet as this would mean exposing company's closed code. It has to be somehow secured. Maybe it's possible to provide a setup when it's protected on networking level, but password protection seems easier (except I can't make it work :sweat_smile: ).

What's interesting is that I can get a proper JWT token with:
```
kuba@***:~$ curl -u <proper user>:<proper password> "https://trow.*****/login"
{"token": "eyJ0eX(...)"}
```

I wonder if building Trow image from the master branch and using it during deployment could solve this problem? Newest version is 0.8.1, whereas the newest available Docker image is from 0.6.4 version.

awoimbee:
Big changes !

I plan on testing and stabilizing this for a while before releasing. Next feature: LRU cache !

Fixes #391
Fixes #145

For #145, I too wanted some sort of CRDT But I just don't have the bandwidth.
Note that there are 2 ways ahead for HA:
1. multiple trows, shared or replicated database, external storage (s3)
2. multiple trows, local database, local storage (federation by Trows querying each other)

I think 2 is more feasible


awoimbee:
I still have some OCI conformance issues, I'm gonna fix these in a following pr

PR #421, opened 2025-03-14
remove duplicate label

explsd:
The `app.kubernetes.io/component: webhooks` label also gets pulling in with `webhook.labels`


awoimbee (approved):
Thanks

PR #422, opened 2025-03-18
add ability to set pod and container security contexts

explsd:
In order to run Trow in my cluster, I require extra security context options set. Normally, I would just by just using Helm postrendering, but it doesn't seem like postrendering works with Helm Chart Hooks. So I've added a way to set the security contexts.

The webhook deployment seems to typically use the same options as the statefulset, so that's what I've done here. However, with the same security context as the statefulset, the webhook deployment cannot create the data files it wants to in the default data-dir directoy. In my own deployment, I just set the data-dir to `/tmp`, but some decision would have to be made before this can be merged in without breaking the webhook deployment.

### DIFF
--- a/charts/trow/values.yaml
+++ b/charts/trow/values.yaml
@@ -9,6 +9,13 @@ image:
   tag:
   pullPolicy: IfNotPresent
 
+podSecurityContext:
+  runAsUser: 333333
+  runAsGroup: 333333
+  fsGroup: 333333
+
+containerSecurityContext: {}
+
 trow:
   ## if using NodePort, this can be set to 127.0.0.1:XXXX
   ## Where "XXXX" is the value of service.nodePort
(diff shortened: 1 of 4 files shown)


charts/trow/values.yaml
```diff
@@ -9,6 +9,13 @@ image:
   tag:
   pullPolicy: IfNotPresent
 
+podSecurityContext:
```
awoimbee, line 12:
```suggestion
## Applies to the Trow Statefulset and the webhooks Deployment
podSecurityContext:
```

awoimbee (approved):
LGTM, thanks

awoimbee:
> However, with the same security context as the statefulset, the webhook deployment cannot create the data files [...]

I'll look into this in a another PR before releasing the chart

